PR #2, opened 2022-07-27
Bootstrap crossplane provider

ZuGao:
## Summary

* Bootstrapping repository
* Change to crossplane provider

## Checklist


operator/generate.go
```diff
@@ -0,0 +1,6 @@
+//go:build generate
+
+// Generate manifests
+//go:generate go run -tags generate sigs.k8s.io/controller-tools/cmd/controller-gen rbac:roleName=manager-role paths=./... output:artifacts:config=../package/rbac
```
ccremer, line 4: Note for future packaging with crossplane: I had to change the path where the RBAC rules are created. See https://github.com/vshn/provider-cloudscale/pull/23
```suggestion
//go:generate go run -tags generate sigs.k8s.io/controller-tools/cmd/controller-gen rbac:roleName=manager-role paths=./... output:artifacts:config=../.kind/rbac
```

package/rbac/role.yaml
```diff
@@ -0,0 +1,64 @@
+---
```
ccremer, line 1: This file should be removed (see my previous comment) as RBAC permissions are handled differently in a package image

ccremer (approved):

Kidswiss (approved):

PR #5, opened 2022-08-04
Add IAM key object

ZuGao:
## Summary
This PR adds support for Exoscale IAM Key:
* Adds IAM APIs;
* Adds IAM controller;
* Adds business logic;
* Adds webhook for CRD validation;
* Adds Crossplane package;
* Configures the local environment.

### Insights
* The IAM key in exoscale is immutable which means there are no API available to update a key.
* The key secret is available only once, at creation time. This makes impossible to ensure the API k8s secret's key secret. Thus, if the key secret is lost from the API k8s secret (manually updated/deleted) the IAM key is lost and cannot be used anymore.
* From the reason above, the secret ref cannot be update in the IAM key CRD. Also the key name, zone and allowed bucket list.
* During creation of the IAM key CRD, there should be at least one bucket specified otherwise the webhook will not allow the creation of the IAM key in exoscale. At this point in time our API does not allow to create IAM Keys with unrestricted access to all buckets.

### Next steps
* New PR for docs;
* New PR for webhook tests and e2e kuttle tests.

### How to manually test
1) Access exoscale.com and add a supper IAM key (or at least one that has full access over object storage `sos`). Save the key and secret in env variables `EXOSCALE_API_KEY` and `EXOSCALE_API_SECRET` respectively.
2) If you are using project VSHN AG in exoscale then the bucket `bucket.test` already exists and you don't have to create a new bucket, otherwise a bucket named `bucket.test` has to be created.
3) Run `make local-install install-samples`.
4) Check the IAM key `kubectl get iamkey iam-key`.
5) Delete the IAM key  `kubectl delete iamkey iam-key`.
6) Try to update the IAM key's zone, key name, secret refer or allowed bucket list.

## Checklist


Kidswiss (changes requested):

operator/iamkeycontroller/create.go
```diff
@@ -0,0 +1,146 @@
...
+		log.V(1).Info("Created IAM Key in exoscale", "keyID", *exoscaleIAM.Key, "displayName", *exoscaleIAM.Name, "tags", exoscaleIAM.Tags)
+		p.exoscaleIAMKey = exoscaleIAM
+		return nil
+	}
+}
+
+func getBuckets(iamResources []v2.IAMAccessKeyResource) string {
+	var buckets strings.Builder
+	for _, iamResource := range iamResources {
+		buckets.WriteString(",")
```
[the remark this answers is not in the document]
ZuGao: We can add an `if` but take in consideration that the webhook does not allow to create iam keys with no buckets thus no resources.

operator/iamkeycontroller/webhook.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+		if newIAMKey.Spec.WriteConnectionSecretToReference.Namespace != oldIAMKey.Spec.WriteConnectionSecretToReference.Namespace {
+			return fmt.Errorf("an IAMKey named %q has been created already, you cannot change the connection secret namespace",
+				oldIAMKey.Status.AtProvider.KeyName)
+		}
+	}
+	return nil
+}
+
+func stringArrayEquals(a []string, b []string) bool {
```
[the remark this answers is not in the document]
ZuGao: I read somewhere that it is best used in tests and not in production code though I don't find that source anymore, @ccremer do you think I could just use the reflect package here?

operator/iamkeycontroller/create.go
```diff
@@ -0,0 +1,146 @@
...
+		if err != nil {
+			return err
+		}
+		// Limitation by crossplane: The interface managed.ExternalClient doesn't allow updating the resource during creation except annotations.
+		// But we need to somehow store the key ID returned by the creation operation, because exoscale API allows multiple IAM Keys with the same display name.
+		// So we store it in an annotation since that is the only allowed place to update our resource.
+		// However, once we observe the spec again, we will copy the key ID from the annotation to the status field,
+		//  and that will become the authoritative source of truth for future reconciliations.
+		metav1.SetMetaDataAnnotation(&iamKey.ObjectMeta, KeyIDAnnotationKey, *exoscaleIAM.Key)
+		metav1.SetMetaDataAnnotation(&iamKey.ObjectMeta, BucketsAnnotationKey, getBuckets(*exoscaleIAM.Resources))
```
ccremer, line 80: Why do we save the bucket names in the annotations also? AFAICT once we have the KeyID we can look up which bucket names are allowed during `Observe`.

operator/iamkeycontroller/observe.go
```diff
@@ -0,0 +1,117 @@
...
+				pipeline.NewStepFromFunc("check credentials", p.checkSecret),
+			).WithErrorHandler(p.observeCredentialsHandler),
+		),
+	).RunWithContext(ctx)
+	if result.IsFailed() {
+		return managed.ExternalObservation{ResourceExists: true, ResourceUpToDate: false, ConnectionDetails: toConnectionDetails(p.exoscaleIAMKey)}, nil
+	}
+
+	iamKey.Status.AtProvider.KeyName = *p.exoscaleIAMKey.Name
+
```
ccremer, line 62: We should update all possible status entries from p.exoscaleIAMKey

operator/iamkeycontroller/observe.go
```diff
@@ -0,0 +1,117 @@
...
+		if KeyId, exists := iamKey.Annotations[KeyIDAnnotationKey]; exists {
+			iamKey.Status.AtProvider.KeyID = KeyId
+			delete(iamKey.Annotations, KeyIDAnnotationKey)
+			if iamKey.Annotations[BucketsAnnotationKey] != "" {
+				iamKey.Status.AtProvider.Services = exoscalev1.Services{
+					SOS: exoscalev1.SOS{
+						Buckets: strings.Split(iamKey.Annotations[BucketsAnnotationKey], ","),
+					},
+				}
+			}
```
ccremer, line 37: Again, we should retrieve this from the object returned by the exoscale API, not via annotations.

operator/iamkeycontroller/create.go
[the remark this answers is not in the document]
Kidswiss: At a later stage that behavior will change, when we support other services. Wouldn't it be possible then that an IAM key contains other resources as well?

operator/iamkeycontroller/observe.go
```diff
@@ -0,0 +1,117 @@
...
+	log.V(1).Error(err, "Credentials Secret needs reconciling")
+	return nil
+}
+
+func isNotFound(err error) bool {
+	var errResp *url.Error
+	if errors.As(err, &errResp) {
+		return err.(*url.Error).Err.Error() == "resource not found"
+	}
+	return false
```
ccremer, line 116: Is that really necessary? Does this not work?
```suggestion
	return err != nil && err.Error() == "resource not found"
```

operator/iamkeycontroller/pipeline.go
```diff
@@ -0,0 +1,61 @@
...
+func hasSecretRef(iamKey *exoscalev1.IAMKey) pipeline.Predicate {
+	return func(ctx context.Context) bool {
+		return iamKey.Spec.WriteConnectionSecretToReference != nil
+	}
+}
+
+func toConnectionDetails(iamKey *exoscalesdk.IAMAccessKey) managed.ConnectionDetails {
+	return map[string][]byte{
+		exoscalev1.AccessKeyIDName:     []byte(*iamKey.Key),
+		exoscalev1.SecretAccessKeyName: []byte(*iamKey.Secret),
```
ccremer, line 55: AFAIK this only works during `Create`, but not afterwards.
That means that a credentials secret is always required, can't be optional (best check for given name in the webhook). Because without the secret present as a form of cache, we can't tell Crossplane the connection details during normal observations.

operator/iamkeycontroller/update.go
```diff
@@ -0,0 +1,16 @@
...
+	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+)
+
+// Update implements managed.ExternalClient.
+// exoscale.com does not allow any updates on IAM keys.
+func (p *IAMKeyPipeline) Update(ctx context.Context, mg resource.Managed) (managed.ExternalUpdate, error) {
+	log := controllerruntime.LoggerFrom(ctx)
+	log.Info("Updating resource (noop)")
```
ccremer, line 14: Since it's a noop, I would not log it until in debug mode, it would be useless noise.
```suggestion
	log.V(1).Info("Updating resource (noop)")
```

operator/iamkeycontroller/webhook.go
```diff
@@ -0,0 +1,76 @@
...
+
+// ValidateCreate implements admission.CustomValidator.
+func (v *IAMKeyValidator) ValidateCreate(_ context.Context, obj runtime.Object) error {
+	iamKey := obj.(*exoscalev1.IAMKey)
+	v.log.V(1).Info("Validate create", "name", iamKey.Name)
+	if len(iamKey.Spec.ForProvider.Services.SOS.Buckets) == 0 {
+		return fmt.Errorf("an IAMKey named %q should have at least 1 allowed bucket",
+			iamKey.Name)
+	}
+	return nil
```
ccremer, line 25: As mentioned before, I would also check that the secret ref is given.

operator/iamkeycontroller/webhook.go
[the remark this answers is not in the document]
ccremer: I would try to use a K8s-specific comparator, like in https://github.com/kubernetes-sigs/controller-runtime/blob/eb39b8eb28cfe920fa2450eb38f814fc9e8003e8/pkg/controller/controllerutil/controllerutil.go#L215 if possible.
`import k8s.io/apimachinery/pkg/api/equality`

With that, you can maybe skip comparing each individual property and return a single error.

operator/iamkeycontroller/create.go
```diff
@@ -0,0 +1,149 @@
...
+		// IAMKey already exists
+		return managed.ExternalCreation{}, nil
+	}
+
+	pipe := pipeline.NewPipeline().WithBeforeHooks(steps.DebugLogger(ctx)).
+		WithSteps(
+			pipeline.NewStepFromFunc("create IAM key", p.createIAMKeyFn(iam)),
+			pipeline.If(hasSecretRef(iam),
+				pipeline.NewStepFromFunc("create credentials secret", p.createCredentialsSecretFn(iam)),
+			),
```
ccremer, line 39: As argued before, we always need to create the secret
```suggestion
			pipeline.NewStepFromFunc("create credentials secret", p.createCredentialsSecretFn(iam)),
```

operator/iamkeycontroller/observe.go
```diff
@@ -0,0 +1,117 @@
...
+		return managed.ExternalObservation{}, resource.Ignore(isNotFound, err)
+	}
+
+	result := pipeline.NewPipeline().WithBeforeHooks(steps.DebugLogger(ctx)).WithSteps(
+		pipeline.If(hasSecretRef(iamKey),
+			pipeline.NewPipeline().WithNestedSteps("observe credentials secret",
+				pipeline.NewStepFromFunc("fetch credentials secret", p.fetchCredentialsSecretFn(iamKey)),
+				pipeline.NewStepFromFunc("check credentials", p.checkSecret),
+			).WithErrorHandler(p.observeCredentialsHandler),
+		),
```
ccremer, line 55:
```suggestion
		pipeline.NewPipeline().WithNestedSteps("observe credentials secret",
			pipeline.NewStepFromFunc("fetch credentials secret", p.fetchCredentialsSecretFn(iamKey)),
			pipeline.NewStepFromFunc("check credentials", p.checkSecret),
		).WithErrorHandler(p.observeCredentialsHandler),
```

package/crossplane.yaml
```diff
@@ -0,0 +1,16 @@
...
+    meta.crossplane.io/maintainer: VSHN <info@vshn.net>
+    meta.crossplane.io/source: github.com/vshn/provider-exoscale
+    meta.crossplane.io/license: Apache-2
+    meta.crossplane.io/description: |
+      Crossplane provider for managing resources on exoscale.com
+spec:
+  controller:
+    image: ghcr.io/vshn/provider-exoscale/controller:latest
+    permissionRequests: []
+  crossplane:
```
ccremer, line 15: This file is dynamically generated, please put it into .gitignore.

README.md
```diff
@@ -1,126 +1,66 @@
...
+1.  To debug the webhook in an IDE, we need to generate certificates:
+    ```bash
+    make webhook-cert
+    ```
+2.  Start the operator in your IDE with `WEBHOOK_TLS_CERT_DIR` environment set to `.kind`.
+
+3.  Send an admission request sample of the spec:
+    ```bash
+    # send an admission request
+    curl -k -v -H "Content-Type: application/json" --data @samples/admission.k8s.io_admissionreview.json https://localhost:9443/validate-exoscale-crossplane-io-v1-bucket
```
ccremer, line 64: wrong endpoint, the sample is for IAM key.
```suggestion
    curl -k -v -H "Content-Type: application/json" --data @samples/admission.k8s.io_admissionreview.json https://localhost:9443/validate-exoscale-crossplane-io-v1-iamkey
```

operator/iamkeycontroller/create.go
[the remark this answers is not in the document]
ZuGao: It will be possible to add other services but it will not affect the logic for the `sos service`. The only case this will not work is if we allow IAM keys to have access to all buckets thus leaving the allowed list an empty list. From my perspective this is not very desirable as it will create confusion from our client's perspective (why empty list should mean access to all buckets). Going forward with this idea, we actually create and IAM Key with an empty list and the provider would populate in real time all the buckets from into that list but we may have tens if not hundres of buckets in a project which again is not a perfect solution. In any case I updated the code and added the `if` as it adds safety and doesn't have any regression.

test/local.mk
```diff
@@ -0,0 +1,104 @@
...
+.envtest_crd_dir:
+	@mkdir -p $(envtest_crd_dir)
+	@cp -r package/crds $(kind_dir)
+
+.envtest_crds: .envtest_crd_dir
+
+$(kind_dir)/.credentials.yaml:
+	@if [ "$$EXOSCALE_API_KEY" = "" ]; then echo "Environment variable EXOSCALE_API_KEY not set"; exit 1; fi
+	@if [ "$$EXOSCALE_API_SECRET" = "" ]; then echo "Environment variable EXOSCALE_API_SECRET not set"; exit 1; fi
+	kubectl create secret generic --from-literal EXOSCALE_API_KEY=$(shell echo $$EXOSCALE_API_KEY) --from-literal EXOSCALE_API_SECRET=$(shell echo $$EXOSCALE_API_SECRET) -o yaml --dry-run=client api-secret > $@
```
ccremer, line 85: this should avoid leaking the secret to stdout
```suggestion
	kubectl create secret generic --from-literal EXOSCALE_API_KEY=$$EXOSCALE_API_KEY --from-literal EXOSCALE_API_SECRET=$$EXOSCALE_API_SECRET -o yaml --dry-run=client api-secret > $@
```

test/local.mk
```diff
@@ -0,0 +1,104 @@
...
+# for ControllerConfig:
+local-install: export INTERNAL_PACKAGE_IMG = registry.registry-system.svc.cluster.local:5000/$(PROJECT_OWNER)/$(PROJECT_NAME)/package:$(IMG_TAG)
+# for package-push:
+local-install: PACKAGE_IMG = localhost:5000/$(PROJECT_OWNER)/$(PROJECT_NAME)/package:$(IMG_TAG)
+local-install: kind-load-image crossplane-setup registry-setup $(kind_dir)/.credentials.yaml package-push  ## Install Operator in local cluster
+	yq e '.spec.metadata.annotations."local.dev/installed"="$(shell date)"' test/controllerconfig-exoscale.yaml | kubectl apply -f -
+	yq e '.spec.package=strenv(INTERNAL_PACKAGE_IMG)' test/provider-exoscale.yaml | kubectl apply -f -
+	kubectl wait --for condition=Healthy provider.pkg.crossplane.io/provider-exoscale --timeout 60s
+	kubectl -n crossplane-system wait --for condition=Ready $$(kubectl -n crossplane-system get pods -o name -l pkg.crossplane.io/provider=provider-exoscale) --timeout 60s
+	kubectl apply -n crossplane-system -f $(kind_dir)/.credentials.yaml
```
ccremer, line 27: Apply the credentials earlier
```suggestion
	kubectl apply -n crossplane-system -f $(kind_dir)/.credentials.yaml
	yq e '.spec.metadata.annotations."local.dev/installed"="$(shell date)"' test/controllerconfig-exoscale.yaml | kubectl apply -f -
	yq e '.spec.package=strenv(INTERNAL_PACKAGE_IMG)' test/provider-exoscale.yaml | kubectl apply -f -
	kubectl wait --for condition=Healthy provider.pkg.crossplane.io/provider-exoscale --timeout 60s
	kubectl -n crossplane-system wait --for condition=Ready $$(kubectl -n crossplane-system get pods -o name -l pkg.crossplane.io/provider=provider-exoscale) --timeout 60s
```

ccremer (changes requested):

apis/exoscale/v1/iam_types.go
```diff
@@ -0,0 +1,143 @@
...
+	// +kubebuilder:validation:Required
+	// Zone is the name of the zone where the IAM key is created.
+	// The zone must be available in the S3 endpoint.
+	// Cannot be changed after IAMKey is created.
+	Zone string `json:"zone"`
+
+	// +kubebuilder:validation:Required
+	// Services is the exoscale service to which IAMKey gets access to.
+	// Only object storage (sos) service is supported thus the IAMKey will be restricted to access only sos.
+	Services `json:"services,omitempty"`
```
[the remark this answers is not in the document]
ZuGao: We have discussed this topic together with @Kidswiss and we decided that later we will support more services.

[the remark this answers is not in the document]
ccremer: Yes, but my suggestion has nothing to do with adding more services in the `ServicesSpec`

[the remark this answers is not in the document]
ZuGao: No composition but aggregation! Fixed that.

test/local.mk
ZuGao: Good point

operator/iamkeycontroller/webhook.go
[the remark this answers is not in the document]
ZuGao: Went with the k8s equality function.

apis/exoscale/v1/iam_types.go
```diff
@@ -0,0 +1,142 @@
...
+	AccessKeyIDName = "AWS_ACCESS_KEY_ID"
+	// SecretAccessKeyName is the environment variable name for the S3 secret key ("password")
+	SecretAccessKeyName = "AWS_SECRET_ACCESS_KEY"
+)
+
+// SOSSpec is the service type for Object Storage in exoscale
+type SOSSpec struct {
+
+	// +kubebuilder:validation:Required
+	// Buckets is a list of buckets to which IAMKey has access to.
```
ccremer, line 22: I would put in a blank line, so that the kubebuilder comment doesn't become part of the field description.
```suggestion
	// +kubebuilder:validation:Required
	
	// Buckets is a list of buckets to which IAMKey has access to.
```

apis/exoscale/v1/iam_types.go
```diff
@@ -0,0 +1,142 @@
...
+	Buckets []string `json:"buckets,omitempty"`
+}
+
+// ServicesSpec are the accessible exoscale services of the IAMKey.
+type ServicesSpec struct {
+
+	// +kubebuilder:validation:Required
+	// SOSSpec is the Object Storage Service in exoscale.
```
ccremer, line 30:
```suggestion
	// +kubebuilder:validation:Required
	
	// SOSSpec is the Object Storage Service in exoscale.
```

apis/exoscale/v1/iam_types.go
```diff
@@ -0,0 +1,142 @@
...
+
+// IAMKeyParameters are the configurable fields of IAMKey.
+type IAMKeyParameters struct {
+	// KeyName is the name of the Key as presented in the exoscale.com UI.
+	// If empty, the value of `.metadata.annotations."crossplane.io/external-name"` is used.
+	// There can be multiple keys that have the same key name in exoscale.com, but they will have different key IDs.
+	KeyName string `json:"keyName,omitempty"`
+
+	// +kubebuilder:validation:Required
+	// Zone is the name of the zone where the IAM key is created.
```
ccremer, line 42:
```suggestion
	// +kubebuilder:validation:Required
	
	// Zone is the name of the zone where the IAM key is created.
```

apis/exoscale/v1/iam_types.go
```diff
@@ -0,0 +1,142 @@
...
+	// The zone must be available in the S3 endpoint.
+	// Cannot be changed after IAMKey is created.
+	Zone string `json:"zone"`
+
+	// +kubebuilder:validation:Required
+	// Services is the exoscale service to which IAMKey gets access to.
```
ccremer, line 48:
```suggestion
	// +kubebuilder:validation:Required
	
	// Services is the exoscale service to which IAMKey gets access to.
```

operator/iamkeycontroller/webhook.go
```diff
@@ -0,0 +1,53 @@
...
+	v.log.V(1).Info("Validate update")
+
+	if oldIAMKey.Status.AtProvider.KeyID != "" {
+		if !equality.Semantic.DeepEqual(newIAMKey.Spec.ForProvider, oldIAMKey.Spec.ForProvider) {
+			return fmt.Errorf("an IAMKey named %q has been created already, you cannot update it",
+				oldIAMKey.Status.AtProvider.KeyName)
+		}
+		if !equality.Semantic.DeepEqual(newIAMKey.Spec.WriteConnectionSecretToReference, oldIAMKey.Spec.WriteConnectionSecretToReference) {
+			return fmt.Errorf("an IAMKey named %q has been created already, you cannot update the conneciton secret reference",
+				oldIAMKey.Status.AtProvider.KeyName)
```
ccremer, line 42:
```suggestion
			return fmt.Errorf("an IAMKey named %q has been created already, you cannot update the connection secret reference",
				oldIAMKey.Status.AtProvider.KeyName)
```

package/crossplane.yaml
ccremer: This file is still present in the PR...

ccremer (changes requested):
Still some stuff.

Please also have a look at the other 14 suggestions I made.
image

package/crossplane.yaml
ZuGao: I removed it now

apis/exoscale/v1/iam_types.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+	if name := meta.GetExternalName(in); name != "" {
+		return name
+	}
+	return in.Name
+}
+
+func (in *IAMKey) GetZone() string {
+	return in.Spec.ForProvider.Zone
+}
```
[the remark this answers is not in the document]
ZuGao: It's not part of any interface, I just want to make shorter when I get the zone

operator/iamkeycontroller/connector.go
```diff
@@ -0,0 +1,85 @@
...
+
+type IAMKeyConnector struct {
+	kube     client.Client
+	recorder event.Recorder
+
+	iamKey         *exoscalev1.IAMKey
+	providerConfig *providerv1.ProviderConfig
+	apiK8sSecret   *corev1.Secret
+	apiKey         string
+	apiSecret      string
```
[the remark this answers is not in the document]
ZuGao: Changed the logic

operator/iamkeycontroller/create.go
ZuGao: that's a good point, I somehow missed that. I changed the logic and I get the resources from the iamKey in the `observe` step

operator/iamkeycontroller/observe.go
ZuGao: added the buckets.

operator/iamkeycontroller/observe.go
ZuGao: yeah this should suffice

operator/iamkeycontroller/webhook.go
ZuGao: Added the validation during `create` operation.

operator/iamkeycontroller/pipeline.go
ZuGao: Added the constraints in validation webhook

operator/iamkeycontroller/observe.go
ZuGao: No it does not work, I will revert to the previous version.

apis/exoscale/v1/iam_types.go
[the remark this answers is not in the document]
ccremer: In that case I'd rather not have to maintain a func for a simple property retrieval without any logic.

operator/iamkeycontroller/connector.go
```diff
@@ -0,0 +1,92 @@
...
+		pipeline.NewStepFromFunc("read API secret", c.readApiKeyAndSecret),
+	).RunWithContext(ctx)
+	if result.IsFailed() {
+		return nil, result.Err()
+	}
+	exoscaleClient := pipeline.MustLoadFromContext(ctx, exoscaleClientKey{}).(*exoscalesdk.Client)
+	return NewPipeline(c.kube, c.recorder, exoscaleClient), nil
+}
+
+func (c *IAMKeyConnector) fetchProviderConfigFn(iamKey exoscalev1.IAMKey) func(ctx context.Context) error {
```
ccremer, line 56:
```suggestion
func (c *IAMKeyConnector) fetchProviderConfigFn(iamKey *exoscalev1.IAMKey) func(ctx context.Context) error {
```

operator/iamkeycontroller/connector.go
```diff
@@ -0,0 +1,92 @@
...
+
+// Connect implements managed.ExternalConnector.
+func (c *IAMKeyConnector) Connect(ctx context.Context, mg resource.Managed) (managed.ExternalClient, error) {
+	ctx = pipeline.MutableContext(ctx)
+	log := ctrl.LoggerFrom(ctx)
+	log.V(1).Info("Connecting resource")
+
+	iamKey := fromManaged(mg)
+	result := pipeline.NewPipeline().WithBeforeHooks(steps.DebugLogger(ctx)).WithSteps(
+		pipeline.NewStepFromFunc("fetch provider config", c.fetchProviderConfigFn(*iamKey)),
```
ccremer, line 45:
```suggestion
		pipeline.NewStepFromFunc("fetch provider config", c.fetchProviderConfigFn(iamKey)),
```

operator/iamkeycontroller/create.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+
+		exoscaleIAM, err := exoscaleClient.CreateIAMAccessKey(ctx, iamKey.GetZone(), iamKey.GetKeyName(), v2.CreateIAMAccessKeyWithResources(keyResources))
+		if err != nil {
+			return err
+		}
+
+		if err != nil {
+			return err
+		}
```
ccremer, line 71:
```suggestion
		if err != nil {
			return err
		}

```

operator/iamkeycontroller/observe.go
```diff
@@ -0,0 +1,122 @@
...
+func (p *IAMKeyPipeline) checkSecret(_ context.Context) error {
+	data := p.credentialsSecret.Data
+
+	if len(data) == 0 {
+		return fmt.Errorf("secret %q does not have any data", fmt.Sprintf("%s/%s", p.credentialsSecret.Namespace, p.credentialsSecret.Name))
+	}
+
+	// Populate secret key from the secret credentials as exoscale IAM get operation does not return the secret key
+	secret := string(data[exoscalev1.SecretAccessKeyName])
+	p.exoscaleIAMKey.Secret = &secret
```
ccremer, line 93: Is `p.exoscaleIAMKey.Key` not required to be set? (in order for `toConnectionDetails()` to work.

ccremer (approved):
Some small stuff, rest LGTM

operator/iamkeycontroller/observe.go
ZuGao: the key id is set earlier in `getIAMKeyFn`function.

PR #10, opened 2022-08-09
Add documentation

ZuGao:
## Summary

Added documentation:
* Installation guide;
* How to create a release;
* Examples.

## Checklist

### DIFF
--- a/docs/antora-build.mk
+++ b/docs/antora-build.mk
@@ -1,4 +1,5 @@
 docs_out_dir := ./.public
+docs_moduleroot_dir := docs/modules/ROOT
 
 docker_opts ?= --rm --tty --user "$$(id -u)"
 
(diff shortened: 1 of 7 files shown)


docs/antora-build.mk
```diff
@@ -1,4 +1,5 @@
 docs_out_dir := ./.public
+docs_moduleroot_dir := docs/modules/ROOT
```
ccremer, line 2: Why this change? doesn't seem to be used.

Kidswiss:
LGTM

But shouldn't this PR also have the label `documentation`?

Kidswiss (approved):

ZuGao:
> LGTM
> 
> But shouldn't this PR also have the label `documentation`?

yes you are right, updated the label.

docs/antora-build.mk
ZuGao: It is used in [Makefile](https://github.com/vshn/provider-exoscale/blob/master/Makefile#L70). Unfortunately rewriting the commit history is not always perfect.

ccremer (approved):

PR #12, opened 2022-08-12
Add controller for provisioning SOS Buckets

ZuGao:
## Summary

* Implementation of S3 bucket controller (CRUD)
* Refactor code for go-command-pipeline 0.19 version

Next PRs:
* Add tests
* Update documentation
* Mandatory provider config for buckets and iam keys 

Local testing with `DeleteIfEmpty` configuration bucket:
1) `make local-install install-samples`
2) `kubectl get bucket bucket`
3) Check object on exoscale.com
4) `kubectl delete bucket bucket`
5) Check object on exoscale.com

Local testing with `DeleteAll` configuration bucket (s3cmd required):
1) `make local-install install-samples`
2) `kubectl get bucket bucket`
3) Check object on exoscale.com
4)  `kubectl edit bucket bucket` -> ` .spec.forProvider.bucketDeletionPolicy: DeleteAll`
5) `s3cmd --configure` (add your exoscale access and secret keys along with the s3 endpoint which is `sos-ch-gva-2.exo.io`)
6) `s3cmd sync <path to any small folder> s3://bucket.test.1/`
7) `kubectl delete bucket bucket`
8) Check object on exoscale.com

## Checklist


operator/controllerutil/connector_common.go
```diff
@@ -0,0 +1,62 @@
...
+	ProviderConfigName string
+
+	providerConfig   *providerv1.ProviderConfig
+	credentialSecret *corev1.Secret
+}
+
+func (c *GenericConnector) FetchProviderConfig(ctx *GenericConnectContext) error {
+	if ctx.ProviderConfigName == "" {
+		return fmt.Errorf(".spec.providerConfigRef.Name is required")
+	}
```
ccremer, line 39: Why don't we enforce this in the admission webhook for Create/Update operations?

operator/iamkeycontroller/create.go
```diff
@@ -10,11 +10,10 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
 	v2 "github.com/exoscale/egoscale/v2"
```
ccremer, line 11: I wasn't sure what `v2` is, so I'd like to see it a bit more clearly and to be consistent with other usages.
```suggestion
	exoscalesdk "github.com/exoscale/egoscale/v2"
```

package/webhook/manifests.yaml
```diff
@@ -5,6 +5,25 @@ metadata:
...
+      path: /validate-exoscale-crossplane-io-v1-bucket
+  failurePolicy: Fail
+  name: buckets.exoscale.crossplane.io
+  rules:
+  - apiGroups:
+    - exoscale.crossplane.io
+    apiVersions:
+    - v1
+    operations:
+    - UPDATE
```
ccremer, line 23: Don't forget to add CREATE if we check provider config in the webhook.

test/e2e/provider/00-assert.yaml
```diff
@@ -18,7 +18,7 @@ spec:
     services:
       sos:
         buckets:
-          - bucket.test.1
+          - bucket.test.e2e
```
ccremer, line 21: Out of curiousity: Why do we delimit bucket names with `.` instead of `-` as we do in cloudscale?

ccremer (commented):
Some unit tests seem to be missing?

operator/iamkeycontroller/create.go
ZuGao: Good observation, I updated it.

package/webhook/manifests.yaml
ZuGao: Ah yes, I forgot about the provider config validation. I will add it.

test/e2e/provider/00-assert.yaml
ZuGao: I had some issues during development with "-". I thought it's not allowed in exoscale. I did some manual tests now and everything seem to work as expected. I will change from `.` to `-`.

ZuGao:
> Some unit tests seem to be missing?

I will add tests in another PR.

operator/controllerutil/connector_common.go
ZuGao: Updated the logic and moved this to the validation webhook

operator/bucketcontroller/connector.go
```diff
@@ -0,0 +1,119 @@
...
+// Connect implements managed.ExternalConnecter.
+func (c *bucketConnector) Connect(ctx context.Context, mg resource.Managed) (managed.ExternalClient, error) {
+	ctx = pipeline.MutableContext(ctx)
+	log := controllerruntime.LoggerFrom(ctx)
+	log.V(1).Info("Connecting resource")
+
+	bucket := fromManaged(mg)
+
+	if isBucketAlreadyDeleted(bucket) {
+		// We do this to prevent after-deletion reconciliations since there is a chance that we might not have the access credentials anymore.
```
Kidswiss, line 40: This sentence isn't true for the exoscale provider though? We're using the admin credentials from the provider config. So I'd remove this comment.

Not doing a reconciliation loop if the bucket doesn't exist still makes sense though.

Kidswiss (changes requested):

operator/bucketcontroller/connector.go
ZuGao: yes, the comment is miss-leading.

Kidswiss (approved):

operator/bucketcontroller/webhook.go
```diff
@@ -0,0 +1,54 @@
...
+
+// ValidateCreate implements admission.CustomValidator.
+func (v *BucketValidator) ValidateCreate(_ context.Context, obj runtime.Object) error {
+	bucket := obj.(*exoscalev1.Bucket)
+	v.log.V(1).Info("Validate create", "name", bucket.Name)
+
+	providerConfigRef := bucket.Spec.ProviderConfigReference
+	if providerConfigRef == nil || providerConfigRef.Name == "" {
+		return fmt.Errorf("a bucket named %q requires a provider config reference",
+			bucket.Name)
```
ccremer, line 25: The returned error message is slightly ambiguous. Is the name referring to the bucket in the spec or the metadata name? Since the below will be wrapped in a kubernetes-generated error that already has the metadata.name, we don't need that info again.
```suggestion
		return fmt.Errorf(".spec.providerConfigRef.name is required")
```

operator/bucketcontroller/webhook.go
```diff
@@ -0,0 +1,54 @@
...
+	return nil
+}
+
+// ValidateUpdate implements admission.CustomValidator.
+func (v *BucketValidator) ValidateUpdate(_ context.Context, oldObj, newObj runtime.Object) error {
+	newBucket := newObj.(*exoscalev1.Bucket)
+	oldBucket := oldObj.(*exoscalev1.Bucket)
+	v.log.V(1).Info("Validate update")
+
+	if oldBucket.Status.AtProvider.BucketName != "" {
```
ccremer, line 36: We still need to check if the `spec.providerConfigRef` contains a name (like in ValidateCreate()`. Otherwise, the user can create a resource with ref set, but then remove again in an update operation.

operator/iamkeycontroller/webhook.go
```diff
@@ -28,6 +28,12 @@ func (v *IAMKeyValidator) ValidateCreate(_ context.Context, obj runtime.Object)
 		return fmt.Errorf("an IAMKey named %q requires a connection secret reference with name and namespace",
 			iamKey.Name)
 	}
+
+	providerConfigRef := iamKey.Spec.ProviderConfigReference
+	if providerConfigRef == nil || providerConfigRef.Name == "" {
+		return fmt.Errorf("an IAMKey named %q requires a provider config reference",
+			iamKey.Name)
```
ccremer, line 35:
```suggestion
		return fmt.Errorf(".spec.providerConfigRef.name is required")
```

ccremer (commented):
Code LGTM, haven't tested it though.

operator/bucketcontroller/webhook.go
```diff
@@ -0,0 +1,57 @@
...
+				oldBucket.Status.AtProvider.BucketName)
+		}
+		if newBucket.Spec.ForProvider.Zone != oldBucket.Spec.ForProvider.Zone {
+			return fmt.Errorf("a bucket named %q has been created already, you cannot change the zone",
+				oldBucket.Status.AtProvider.BucketName)
+		}
+		providerConfigRef := newBucket.Spec.ProviderConfigReference
+		if providerConfigRef == nil || providerConfigRef.Name == "" {
+			return fmt.Errorf(".spec.providerConfigRef.name is required")
+		}
```
ccremer, line 47: This check needs to be outside of `if oldBucket.Status.AtProvider.BucketName != "" {`

operator/iamkeycontroller/webhook.go
ccremer: This one is still open?

operator/bucketcontroller/webhook.go
ZuGao: yep, updated it.

operator/iamkeycontroller/webhook.go
ZuGao: I missed the iamkey. Updated.

ccremer (approved):

PR #15, opened 2022-08-17
Update webhook integration tests and kuttl e2e tests

ZuGao:
## Summary

* Add bucket webhook tests.
* Adjust existing IAM key webhook tests.
* Update e2e tests.

## Checklist

- [x] Update tests.


operator/bucketcontroller/webhook_test.go
```diff
@@ -0,0 +1,193 @@
...
+	exoscalev1 "github.com/vshn/provider-exoscale/apis/exoscale/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestBucketValidator_ValidateCreate_RequireProviderConfig(t *testing.T) {
+	tests := map[string]struct {
+		providerName  string
+		expectedError string
+	}{
+		"GivenProviderName_ThenExpectMoError": {
```
Kidswiss, line 19:
```suggestion
		"GivenProviderName_ThenExpectNoError": {
```

operator/bucketcontroller/webhook_test.go
```diff
@@ -0,0 +1,193 @@
...
+		})
+	}
+}
+
+func TestBucketValidator_ValidateUpdate_RequireProviderConfig(t *testing.T) {
+	tests := map[string]struct {
+		providerName  string
+		expectedError string
+	}{
+		"GivenProviderName_ThenExpectMoError": {
```
Kidswiss, line 110:
```suggestion
		"GivenProviderName_ThenExpectNoError": {
```

Kidswiss (changes requested):
There are some typos.

operator/iamkeycontroller/webhook_test.go
```diff
@@ -11,165 +11,150 @@ import (
...
+				require.NoError(t, err)
+			}
+		})
+	}
+}
+
+func TestIAMKeyValidator_ValidateCreate_RequireWriteConnectionSecretToRef(t *testing.T) {
+	tests := map[string]struct {
+		iamKey                exoscalev1.IAMKey
+		connectionSecretToRef xpv1.SecretReference
```
ccremer, line 62: This should be a pointer, since the spec allows `nil`.

Also, currently there isn't a test case that tests that `WriteConnectionSecretToRef` is nil.
```suggestion
		connectionSecretToRef *xpv1.SecretReference
```

operator/iamkeycontroller/webhook_test.go
```diff
@@ -11,165 +11,150 @@ import (
...
+				require.NoError(t, err)
+			}
+		})
+	}
+}
+
+func TestIAMKeyValidator_ValidateCreate_RequireProviderConfigToRef(t *testing.T) {
+	tests := map[string]struct {
+		iamKey              exoscalev1.IAMKey
+		providerConfigToRef xpv1.Reference
```
ccremer, line 118: Also here, please add a test case to check that the `nil` is handled correctly as well.
```suggestion
		providerConfigToRef *xpv1.Reference
```

operator/iamkeycontroller/webhook_test.go
```diff
@@ -179,229 +164,193 @@ func TestIAMKeyValidator_ValidateCreate(t *testing.T) {
 	}
 }
 
-func TestIAMKeyValidator_ValidateUpdate(t *testing.T) {
+func TestIAMKeyValidator_ValidateUpdate_RequireProviderConfigToRef(t *testing.T) {
 	tests := map[string]struct {
-		newIAMKey     exoscalev1.IAMKey
-		oldIAMKey     exoscalev1.IAMKey
-		expectedError string
+		providerConfigToRef xpv1.Reference
```
ccremer, line 169: Same here
```suggestion
		providerConfigToRef *xpv1.Reference
```

ccremer (changes requested):

operator/bucketcontroller/webhook_test.go
```diff
@@ -0,0 +1,193 @@
...
+		expectedError string
+	}{
+		"GivenProviderName_ThenExpectNoError": {
+			providerName: "provider-config",
+		},
+		"GivenNoProviderName_ThenExpectError": {
+			providerName:  "",
+			expectedError: `.spec.providerConfigRef.name is required`,
+		},
+	}
```
ccremer, line 117: Sorry for not spotting it earlier, but the provider config reference can also be possibly nil and we should check for that.

Alternatively, I think it may start making sense to extract some individual checks into their own function, so that we don't have to test them twice (once in Create and once in Update)

ZuGao: In this case it makes sense to group them by object(only 1 or 2 strings) as long as the object is not complex.

ccremer (approved):

PR #18, opened 2022-08-26
Fix IAMkeys allowed operations

ZuGao:
## Summary

* Added the list of bucket operations that the IAMKeys should have access to.
* Update the docs

Currently there is a bug in the Exoscale backend where several `list-*` operations are not working as intended. As soon as the S3 provider will fix the issue I will add the remaining operations.

## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/explanations/iamkey.adoc
+++ b/docs/modules/ROOT/pages/explanations/iamkey.adoc
@@ -1,0 +1,20 @@
+== IAMKeys permissions
+
+The IAMKeys have restricted access to exoscale API.
+The Object Storage `sos` service is the only service the keys have access to.
+The following are the allowed operations on buckets:
+
+- abort-sos-multipart-upload
+- delete-sos-object
+- get-sos-bucket-acl
+- get-sos-bucket-cors
+- get-sos-bucket-location
+- get-sos-object
+- get-sos-object-acl
+- get-sos-presigned-url
+- list-sos-bucket
+- list-sos-bucket-multipart-uploads
+- put-sos-bucket-acl
+- put-sos-bucket-cors
+- put-sos-object
+- put-sos-object-acl
--- a/docs/modules/ROOT/nav.adoc
+++ b/docs/modules/ROOT/nav.adoc
@@ -12,6 +12,7 @@
 //* xref:references/example.adoc[Example Reference]
 
 .Explanation
+* xref:explanations/iamkey.adoc[IAMKeys permissions]
 * For Developers
 ** xref:explanations/dev/iamkey-reconciliation.adoc[ObjectsUser Reconciliation]
 ** xref:explanations/dev/bucket-reconciliation.adoc[Bucket Reconciliation]
--- a/operator/iamkeycontroller/create.go
+++ b/operator/iamkeycontroller/create.go
@@ -61,7 +61,18 @@ func (p *IAMKeyPipeline) createIAMKey(ctx *pipelineContext) error {
 		keyResources = append(keyResources, keyResource)
 	}
 
-	exoscaleIAM, err := p.exoscaleClient.CreateIAMAccessKey(ctx, iamKey.Spec.ForProvider.Zone, iamKey.GetKeyName(), exoscalesdk.CreateIAMAccessKeyWithResources(keyResources))
+	zone := iamKey.Spec.ForProvider.Zone
+	iamKeyName := iamKey.GetKeyName()
+	iamKeyOpts := []exoscalesdk.CreateIAMAccessKeyOpt{
+		// Allowed object storage operations on the new IAM key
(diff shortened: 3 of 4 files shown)


docs/modules/ROOT/pages/explanations/iamkey.adoc
```diff
@@ -0,0 +1,19 @@
+== IAMKeys permissions
+
+The IAMkeys have restricted access to exoscale API.
+The Object Storage `sos` service is the only service the keys have access to.
+The following are the allowed operations on buckets:
+
+- abort-sos-multipart-upload
+- delete-sos-bucket
```
ccremer, line 8: We explictly removed the permission to create buckets. Shouldn't we also remove the `delete-sos-bucket`? The bucket is being deleted with the secret as defined in the provider config, not the credentials of the IAM key itself.

ZuGao: That's a good point you raised. In normal conditions the client would not know the names of buckets it does not have access to which may be a valid point to leave the deletion in there but we also put our provider in difficult situation when the bucket suddenly disappears while it tries to reconcile the CRD.  @Kidswiss what's your take on that?

Kidswiss: @ZuGao Do you know how the provider would react if the bucket is deleted? Would it try to re-create it, or simply error out?

If it's the latter, I'd remove the permission to delete, as it could lead to deadlocks for a customer.

ZuGao: Actually the bucket is being recreated without any issues :)

ccremer: I don't follow what you discuss now. This is about the IAM key having more permissions than it needs to. Bucket deletion and creation happens with a different credentials.

ZuGao: It does not matter which credentials the provider is using. We are trying to figure out what happens if the user would delete its own buckets outside the provider control with the IAM key provided to him, for instance with an s3 client. How would the provider react? Well I figured out that the provider would recreate it with no issues, so we could actually leave the delete in place but nevertheless I believe we should remove this operation as you suggested. We should leave the deletion of buckets only via the CRD manifest or via its claim.

Kidswiss: Fine by me.

ZuGao: I updated the operation list.

ccremer: Thanks for the explanation. In that case I would opt to allow creation and deletion only via claim to have exactly 1 supported way of creating and deleting buckets.

ccremer: > I updated the operation list.

Not in the docs :)

ZuGao: yeah... fixed it. Do I leave this PR open until Exoscale fixes the issue?

ccremer: > until Exoscale fixes the issue?

What issue?

ZuGao: I wrote in the description about a bug in exoscale.

ccremer: ah, I missed/forgot that one :)

For me it's fine to merge, we probably don't know how long it takes for the issue to be fixed. I wouldn't keep this open for weeks and the remaining operations can be re-added later as well.

docs/modules/ROOT/pages/explanations/iamkey.adoc
```diff
@@ -0,0 +1,18 @@
+== IAMKeys permissions
+
```
ccremer, line 2: Missing nav link :)

ccremer (approved):

docs/modules/ROOT/pages/explanations/iamkey.adoc
ZuGao: I will open an issue in GitHub then so we have a ticket in our system. Please approve it so I can merge it

docs/modules/ROOT/pages/explanations/iamkey.adoc
ZuGao: done, thank you.

docs/modules/ROOT/pages/explanations/iamkey.adoc
ZuGao: Opened [issue]( https://github.com/vshn/provider-exoscale/issues/19)

Kidswiss (approved):

PR #35, opened 2022-09-30
Define API scheme for managed PostgreSQL DBaaS

ccremer:
## Summary

* Define API for Postgresql

Note: The controller is set up in the follow-up PR: #37 

## Checklist

- [x] Update tests.
- [ ] ~I have run `make test-e2e` and e2e tests pass successfully~


apis/exoscale/v1/dbaas_types.go
```diff
@@ -0,0 +1,138 @@
...
+	Message string `json:"message,omitempty"`
+	// Type of the notification.
+	Type exoscaleoapi.DbaasServiceNotificationType `json:"type,omitempty"`
+	// Metadata contains additional data.
+	Metadata runtime.RawExtension `json:"metadata,omitempty"`
+}
+
+// MaintenanceSpec contains settings to control the maintenance of an instance.
+type MaintenanceSpec struct {
+	// DayOfWeek specifies at which weekday the maintenance is held place.
```
Kidswiss, line 39: Maybe add the comment for kubebuilder to handle this as an enum?

operator/pipelineutil/exoscale_connector.go
```diff
@@ -0,0 +1,99 @@
...
+	providerConfig   *providerv1.ProviderConfig
+	credentialSecret *corev1.Secret
+	apiKey           string
+	apiSecret        string
+	exoscaleClient   *exoscalesdk.Client
+	opts             []exoscalesdk.ClientOpt
+}
+
+// OpenExoscaleClient fetches the ProviderConfig given by the name, fetches the API secret and returns a ExoscaleConnector with an initialized client.
+func OpenExoscaleClient(ctx context.Context, kube client.Client, providerConfigRef string, opts ...exoscalesdk.ClientOpt) (*ExoscaleConnector, error) {
```
Kidswiss, line 43: This should probably go in the follow-up PR?

Kidswiss (changes requested):
Just some nitpicking, otherwise LGTM

operator/pipelineutil/exoscale_connector.go
ccremer: I added a commit that refactors some code. but I _may_ be able to reorder the commits

apis/exoscale/v1/dbaas_types.go
ccremer: At first I wanted to avoid having a mix of kubebuilder validation and webhooks. I'd rather have only 1 place for validation. But for enums the validation code is actually a pain to have, e.g. using switch cases. So yeah, I can add that.

Kidswiss: TBH I'd go the path of least resistance. If there's a very complicated verification -> webhook. If it can be handled built-in provided by kubebuilder -> kubebuilder. No need to re-invent the wheel...

Kidswiss (approved):

apis/exoscale/v1/dbaas_types.go
```diff
@@ -0,0 +1,155 @@
...
+	// Allowed values are [monday, tuesday, wednesday, thursday, friday, saturday, sunday, never]
+	DayOfWeek exoscaleoapi.DbaasServiceMaintenanceDow `json:"dayOfWeek,omitempty"`
+
+	// TimeOfDay for installing updates in UTC.
+	// Format: "hh:mm:ss".
+	TimeOfDay TimeOfDay `json:"timeOfDay,omitempty"`
+}
+
+var timeOfDayRegex = regexp.MustCompile("^([0-2]?[0-3]):([0-5][0-9]):([0-5][0-9])$")
```
ZuGao, line 73: This regex patter does not compile values for hour 4 - 9 (also 04 - 09). Maybe you could try `^(2[0-3]|[01]?[0-9]):([0-5][0-9]):([0-5][0-9])$`.

apis/exoscale/v1/dbaas_types.go
```diff
@@ -0,0 +1,155 @@
...
+// Format: "hh:mm:ss".
+type TimeOfDay string
+
+// String implements fmt.Stringer.
+func (t TimeOfDay) String() string {
+	return string(t)
+}
+
+// NewTimeOfDay returns a new TimeOfDay acceptable string.
+func NewTimeOfDay(v string) (TimeOfDay, error) {
```
ZuGao, line 87: I suppose this function will be used in a next PR?

apis/exoscale/v1/postgresql_types.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// PostgreSQLStatus represents the observed state of a PostgreSQL.
+type PostgreSQLStatus struct {
+	xpv1.ResourceStatus `json:",inline"`
+	AtProvider          PostgreSQLObservation `json:"atProvider,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:printcolumn:name="State",type="string",JSONPath=".status.conditions[?(@.type=='Ready')].reason"
```
ZuGao, line 57: This line should probably be deleted.

ZuGao (changes requested):
Overall looks great. Just made some remarks and probably there's a problem with regex.

apis/exoscale/v1/postgresql_types.go
ccremer: I don't see why?

apis/exoscale/v1/dbaas_types.go
ccremer: It's actually not. Cleaning up

apis/exoscale/v1/dbaas_types.go
ccremer: how could I be so dumb :see_no_evil: good catch!

apis/exoscale/v1/postgresql_types.go
ZuGao: Isn't type `Ready` already configured with the next line?

ccremer: The column "Ready" displays the `.status` field of the Ready condtion, not the `.reason`. The value is given when the Ready is actually `False` and "State" would then display the VM state, e.g. `Rebuilding`, `PoweredOff` etc.

ZuGao: Good, thanks for insights.

ZuGao (approved):

PR #37, opened 2022-10-06
Implement controller for managed PostgreSQL DBaaS

ccremer:
## Summary

* Adds a controller for the newly added PostgreSQL API
* Also adds a basic webhook for validating the spec.

Follows #35 

## Checklist

- [x] Update tests.
- [x] Ensure `pgSettings` are passed through correctly and can be compared in the status
- [ ] Create issue to document the available settings for `pgSettings`
- [ ] ~I have run `make test-e2e` and e2e tests pass successfully~ see #38


operator/bucketcontroller/connector.go
```diff
@@ -41,48 +38,29 @@ func (c *bucketConnector) Connect(ctx context.Context, mg resource.Managed) (man
 		return &NoopClient{}, nil
 	}
 
-	pctx := &connectContext{
-		GenericConnectContext: controllerutil.GenericConnectContext{
-			Context:            ctx,
-			ProviderConfigName: bucket.GetProviderConfigName(),
-		},
-		EndpointURL: bucket.Spec.ForProvider.EndpointURL,
+	exo, err := pipelineutil.OpenExoscaleClient(ctx, c.Kube, bucket.GetProviderConfigName())
```
ZuGao, line 41: I expect this function to be called very often in the near future with all the services we will support from Exoscale. Maybe it is a good idea to cash the client for a short amount of time?

operator/mapper/postgresql.go
```diff
@@ -0,0 +1,135 @@
...
+
+	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
+	"github.com/exoscale/egoscale/v2/oapi"
+	"github.com/minio/minio-go/v7/pkg/set"
+	exoscalev1 "github.com/vshn/provider-exoscale/apis/exoscale/v1"
+	"k8s.io/utils/pointer"
+)
+
+// PostgreSQLMapper is a mapper to convert exoscale's SDK resources to Kubernetes API specs.
+type PostgreSQLMapper struct{}
```
ZuGao, line 16: Any reason to create an empty struct here? Any reason to not use simple functions for conversion?

operator/bucketcontroller/connector.go
ccremer: Not worth the complexity IMO. controller-runtime already caches objects when first retrieving (e.g. Secrets, providerconfig). Creating the client itself is inexpensive as well, as there's no "login" process. The heavyweight operations are still the API requests and those can hardly be cached.

operator/mapper/postgresql.go
ccremer: Mainly Naming :tm: I expect that there will be other mappers in the future, e.g. Mysql, Kafka etc. So I didn't want to end up with long function names like `ToPostgreSQLSpec(...)` or `ToPostgreSQLConnectionDetails(...)`.

PR #42, opened 2022-10-17
Add "endpoint" field to bucket

davidgubler:
Due to apparent limitations of Crossplane compositions we need both an "endpoint" and an "endpointURL" field. "endpoint" will contain the simple endpoint host name, while endpointURL will contain the full URL (https://...).

Context: https://ticket.vshn.net/browse/APPCAT-226


Kidswiss (approved):
LGTM

Seems more sensible if we don't have to tell the provider-exoscale where the exoscale endpoint is :D

ccremer (approved):

PR #38, opened 2022-10-11
Add e2e test case for PostgreSQL

ccremer:
## Summary

* Adds e2e test for #37 

## Checklist

- [x] Categorize the PR by setting a good title and adding one of the labels:
      `bug`, `enhancement`, `documentation`, `change`, `breaking`, `dependency`
      as they show up in the changelog
- [x] Update tests.
- [x] Link this PR to related issues.
- [x] I have run `make test-e2e` and e2e tests pass successfully

PR #43, opened 2022-10-19
Implement Updating of PostgreSQL instances

ccremer:
## Summary

* Implements the "Update" part of the API. Users can now update instances based on the spec
* The spec prevents modification of an instance of certain properties if the instance has been provisioned once.

Following properties are immutable by Exoscale API:
* `zone`
* `version` 

## Checklist

- [x] Update tests.
- [ ] ~I have run `make test-e2e` and e2e tests pass successfully~ See #38

### DIFF
--- a/operator/mapper/postgresql.go
+++ b/operator/mapper/postgresql.go
@@ -17,10 +17,14 @@ import (
 // PostgreSQLMapper is a mapper to convert exoscale's SDK resources to Kubernetes API specs.
 type PostgreSQLMapper struct{}
 
-// FromSpec places the given spec into the request body.
-func (PostgreSQLMapper) FromSpec(spec exoscalev1.PostgreSQLParameters, into *oapi.CreateDbaasServicePgJSONBody) error {
+// FromSpecToCreateBody places the given spec into the request body.
+func (PostgreSQLMapper) FromSpecToCreateBody(spec exoscalev1.PostgreSQLParameters, into *oapi.CreateDbaasServicePgJSONBody) error {
+	/**
+	NOTE: If you change anything below, also update FromSpecToUpdateBody().
+	Unfortunately the generated openapi-types in exoscale are unusable for reusing same properties.
+	*/
 	backupSchedule, backupErr := toBackupSchedule(spec.Backup.TimeOfDay)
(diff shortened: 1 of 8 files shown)


operator/mapper/postgresql.go
```diff
@@ -37,6 +41,30 @@ func (PostgreSQLMapper) FromSpec(spec exoscalev1.PostgreSQLParameters, into *oap
 	return nil
 }
 
+// FromSpecToUpdateBody places the given spec into the request body.
+func (PostgreSQLMapper) FromSpecToUpdateBody(spec exoscalev1.PostgreSQLParameters, into *oapi.UpdateDbaasServicePgJSONRequestBody) error {
+	/**
+	NOTE: If you change anything below, also update FromSpecToCreateBody().
```
ZuGao, line 47: Have you evaluated to use reflection in this case?

ZuGao (changes requested):
Can you add in the docs which attributes of the PostgreSQL are updated-able  and which are not?

operator/mapper/postgresql.go
ccremer: I considered several different approaches... new types with interfaces, reflection, some structural changes... eventually I concluded that it doesn't reduce the code and only makes things more complicated. So I left it as is, can still be improved later :shrug:

ccremer:
> Can you add in the docs which attributes of the PostgreSQL are updated-able and which are not?

at first I thought this would be something for the end-user documentation. But maybe it makes sense to document some details. I'll do it in a follow-up PR though.

ZuGao:
Also I noticed whenever an ipFilter is added or changed I get this error:
``` message: 'update failed: cannot update instance: Put "https://api-ch-dk-2.exoscale.com/v2/dbaas-postgres/postgresql-local-dev":
        invalid request: Invalid ''user_config'' ip_filter value ''12.32.143.12/3'''
```

ccremer:
> 12.32.143.12/3

This is an invalid CIDR range, no wonder the API returns you this :)

ZuGao:
I did a dumb test, you are right. Updating the ip filter will take a while to reflect it in the CRD but it's fine.

ZuGao (approved):

PR #50, opened 2022-10-27
Introduce Managed Redis DBaaS

mweibel:
## Summary

* Introduces API (CRD) for Exoscale Redis service
* Adds controller
* Adds e2e-test

## Checklist

- [x] Update tests.


operator/mapper/redis.go
```diff
@@ -0,0 +1,47 @@
+package mapper
```
mweibel, line 1: Warning, personal opinion ahead:

I don't like the `type PostgreSQLMapper struct{}` and associated method receivers pattern. It strikes me as a way to use classes in a language without classes. 

I also don't see why mappers specific to certain controllers should be in a shared package. They don't interest outside of that specific package.

I made a compromise here: I didn't use the empty struct pattern but moved the map functions into the `mapper` package, even though I'd really prefer them to be in the `rediscontroller` package.

operator/rediscontroller/controller_test.go
```diff
@@ -0,0 +1,120 @@
+package rediscontroller
```
mweibel, line 1: This is a preliminary setup of how we could unit test the controllers. I'm not sure how much worth they are but my goal is to have a middleground between e2e tests and manual tests. This way we could test `Observe()`, `Create()`, `Update()`, and `Delete()` better without the need to have a complete crossplane setup. 

Of course the whole `httptest.Server` thing could be improved a lot, this is just a start and a RFC.

operator/rediscontroller/pipeline.go
```diff
@@ -0,0 +1,226 @@
...
+	"context"
+	"fmt"
+	"net/url"
+
+	"github.com/crossplane/crossplane-runtime/pkg/errors"
+	"github.com/crossplane/crossplane-runtime/pkg/event"
+	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+	exoscalesdk "github.com/exoscale/egoscale/v2"
+	"github.com/exoscale/egoscale/v2/api"
```
ccremer, line 13: I would probably introduce an alias to not confuse with kubernetes api
```suggestion
	exoscaleapi "github.com/exoscale/egoscale/v2/api"
```

operator/rediscontroller/webhook_test.go
```diff
@@ -0,0 +1,46 @@
...
+		givenSpec     string
+		expectedError string
+	}{
+		"EmptySettings":     {givenSpec: "", expectedError: ""},
+		"EmptyObject":       {givenSpec: "{}", expectedError: ""},
+		"Null":              {givenSpec: `null`, expectedError: ""},
+		"Float":             {givenSpec: `{"number":0.5}`, expectedError: ""},
+		"Zero":              {givenSpec: `{"number":0}`, expectedError: ""},
+		"Integer":           {givenSpec: `{"number":1}`, expectedError: ""},
+		"NegativeNumber":    {givenSpec: `{"number":-1.5}`, expectedError: ""},
```
ccremer, line 22: Maybe we can extract the core function to parse properties to its own function so that we only need to test this once?

ccremer (commented):
First round of review, didn't go into everything yet.

operator/rediscontroller/pipeline.go
```diff
@@ -0,0 +1,226 @@
...
+// newPipeline returns a new instance of pipeline.
+func newPipeline(client client.Client, recorder event.Recorder, exoscaleClient *exoscalesdk.Client) *pipeline {
+	return &pipeline{
+		kube:     client,
+		recorder: recorder,
+		exo:      exoscaleClient,
+	}
+}
+
+func (p pipeline) Observe(ctx context.Context, mg resource.Managed) (managed.ExternalObservation, error) {
```
ZuGao, line 38: I like this approach of keeping the methods inside one file though at this point I would prefer we keep the structure consistent that is a method per file. We can always decide to do some refactoring later on.

operator/rediscontroller/controller_test.go
ZuGao: That's a good idea to start to have some kind of unit tests in the main methods. We generally skipped this for quite some time but we planned in the future to add such tests.

operator/mapper/redis.go
mweibel: > well, the associated method receivers pattern is actually the closest thing to a class. Look at interface implementations for example. Whether they're called structs or classes makes little difference for a Dev, so I'm fine with that.

mh yeah, that's where we disagree I guess ;)

>  I don't like static functions that have repeatedly the same name over and over again in the function names just to differentiate between other, similar functions.

I believe that shows a structural problem with the current code. The `mappers` package is fine to have, but the `postgres` (or redis for that matter) specific code does IMO not belong there. It's of no use to other packages since it's highly specific to the controller in question. 

>  At that time I was expecting that most types or aliases can be shared between different DB services, so I prepared the mapper package as a separate translation layer between Exoscale API and our provider API.

Ok that's what I thought. 
I think it's fine to have it, but the things we can share only. I.e. postgres stuff should not belong there IMO. I'll be moving the functions for redis back and we can refactor postgres in a separate PR if necessary.

operator/rediscontroller/controller_test.go
mweibel: envtest: yeah saw it and didn't see it in use so I didn't use it :) But yeah - that's fine to me, I'll refactor this accordingly :+1: 

I'll add a few tests for redis then and we can reuse the pattern/suite for other tests in the future.

apis/exoscale/v1/postgresql_types.go
```diff
@@ -46,10 +38,12 @@ type PostgreSQLSpec struct {
...
 	DBaaSParameters `json:",inline"`
-	Maintenance     MaintenanceSpec      `json:"maintenance,omitempty"`
-	Backup          BackupSpec           `json:"backup,omitempty"`
-	NoteStates      []NodeState          `json:"noteStates,omitempty"`
-	PGSettings      runtime.RawExtension `json:"pgSettings,omitempty"`
+	// Version is the (major) version identifier for the instance.
+	Version     string               `json:"version,omitempty"`
+	Maintenance MaintenanceSpec      `json:"maintenance,omitempty"`
+	Backup      BackupSpec           `json:"backup,omitempty"`
+	NoteStates  []NodeState          `json:"noteStates,omitempty"`
```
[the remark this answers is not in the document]
mweibel: yeah saw that. However that changes the CRD in a meaningful way. Are we still able to do that at this point?

operator/rediscontroller/pipeline.go
mweibel: ok, fair enough.

operator/mapper/redis.go
ZuGao: > If you think that doesn't really apply, I'm fine with moving these mappings to their DB service-specific packages and that would also not "require" empty structs due to the naming problem.

Actually that would solve the naming problem if we move the mapper inside it's own package. We cannot reuse the mappers due to the oapi design and that's a pity because a lot of the db services are having the same attributes.

ccremer: yeah, let's move the DB specific mappers to the controller packages :+1:

apis/exoscale/v1/postgresql_types.go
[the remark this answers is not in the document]
ccremer: yes, we should be able to do that

operator/mapper/redis.go
```diff
@@ -0,0 +1,47 @@
...
+	exoscalev1 "github.com/vshn/provider-exoscale/apis/exoscale/v1"
+	"k8s.io/utils/pointer"
+)
+
+func MapRedisObservation(log logr.Logger, instance oapi.DbaasServiceRedis) exoscalev1.RedisObservation {
+	notifications, err := toNotifications(instance.Notifications)
+	if err != nil {
+		// just ignore those if they fail to map
+		log.Error(err, "unable to map notifications", "notifications", instance.Notifications)
+	}
```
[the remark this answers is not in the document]
mweibel: `MapRedisParameters` is a bit different since not being able to parse settings seems crucial to me, while not being able to parse notifications not so much. The difference is that `MapRedisObservation` would still succeed with the rest mapped even if notifications is not working. 

But I can still return a specific error which we can just log.

operator/mapper/redis.go
mweibel: will do in a separate PR for postgres :+1:

operator/rediscontroller/observe.go
```diff
@@ -0,0 +1,170 @@
...
+	if err != nil {
+		return managed.ExternalObservation{}, fmt.Errorf("unable to parse connection details: %w", err)
+	}
+
+	observation := managed.ExternalObservation{
+		ResourceExists:          true,
+		ResourceUpToDate:        isUpToDate(&redisInstance.Spec.ForProvider, rp, log),
+		ResourceLateInitialized: false,
+		ConnectionDetails:       cd,
+		Diff:                    "",
```
ccremer, line 81:
```suggestion
		ConnectionDetails:       cd,
```

operator/rediscontroller/controller_test.go
```diff
@@ -0,0 +1,296 @@
...
+		p:        ts.p,
+	}, 0)
+	ts.Require().NoError(err)
+}
+
+func (ts *RedisControllerTestSuite) TestHappyPath() {
+	settings := map[string]any{
+		"maxmemory_policy": "noeviction",
+	}
+	mg := newRedisInstance("test", settings)
```
ccremer, line 60: This seems like a rather large test case that tests everything :tm: . This makes maintaining the test very hard IMO.
Is there no chance to test individual actions in isolation of eachother? Like
- Test if creation works: start with nothing, end up with a "running" instance
- Test if update works: start with an "existing" instance, update a field, await status updates
- Test if delete works: start with "existing" instance, delete it. assert nothing exists.

To me, create, update and delete are three distinct operations worth their own test case. In case "creation" of an instance fails, it shouldn't fail the update test case as well, for example. Or deletion.

apis/exoscale/v1/redis_types.go
```diff
@@ -0,0 +1,112 @@
...
+	xpv1.ResourceStatus `json:",inline"`
+	AtProvider          RedisObservation `json:"atProvider,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:printcolumn:name="State",type="string",JSONPath=".status.conditions[?(@.type=='Ready')].reason"
+// +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=".status.conditions[?(@.type=='Ready')].status"
+// +kubebuilder:printcolumn:name="Synced",type="string",JSONPath=".status.conditions[?(@.type=='Synced')].status"
+// +kubebuilder:printcolumn:name="External Name",type="string",JSONPath=".metadata.annotations.crossplane\\.io/external-name"
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
```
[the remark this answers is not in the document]
mweibel: that might make sense for postgres too, no?

internal/operatortest/generate.go
```diff
@@ -0,0 +1,4 @@
+package operatortest
```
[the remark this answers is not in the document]
mweibel: I believe that's not really necessary but to keep consistent I added it :+1:

test/e2e/redis/00-assert.yaml
```diff
@@ -0,0 +1,56 @@
...
+  writeConnectionSecretToRef:
+    name: e2e-test-redis-details
+    namespace: default
+status:
+  atProvider:
+    nodeStates:
+      - name: e2e-test-redis-1
+        role: master
+        state: running
+    version: 7.0.5
```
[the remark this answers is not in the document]
mweibel: good point :+1:

operator/rediscontroller/observe.go
```diff
@@ -0,0 +1,170 @@
...
+	return observation, nil
+}
+
+func isUpToDate(current, external *exoscalev1.RedisParameters, log logr.Logger) bool {
+	if external == nil {
+		return false
+	}
+	extIPFilter := []string(external.IPFilter)
+	checks := map[string]bool{
+		"IPFilter":              mapper.IsSameStringSet(current.IPFilter, &extIPFilter),
```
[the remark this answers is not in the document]
mweibel: it's also used by postgres to check for equality. There the types differ a bit and I believe we should refactor this afterwards to make it consistent. Agree?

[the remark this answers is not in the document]
ccremer: fine by me

operator/rediscontroller/controller_test.go
mweibel: Yeah, that does make sense. However, for the non-creation cases, I'd need to still do the creation regardless since I (correct me if I'm wrong) can't add objects just like that in the envtest environment. 

I can certainly add this, but not sure how much it improves the code TBH.

ccremer: > can't add objects just like that in the envtest environment.

sure you can, you just need to prepare the spec and update status accordingly, so that the reconciler thinks it's already created from a previous "loop".

operator/postgresqlcontroller/observe.go
```diff
@@ -35,7 +35,7 @@ func (p *Pipeline) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 	isUpToDate := mpr.IsResourceUpToDate(pgInstance, resp.JSON200)
 	setConditionFromState(pgExo, pgInstance)
 
-	ca, err := p.exo.GetDatabaseCACertificate(ctx, pgInstance.Spec.ForProvider.Zone)
+	ca, err := p.exo.GetDatabaseCACertificate(ctx, string(pgInstance.Spec.ForProvider.Zone))
```
ccremer, line 38: I suggest to add `func (z Zone) String() string` to the Zone, so that it implements fmt.Stringer and we can do this instead:
```suggestion
	ca, err := p.exo.GetDatabaseCACertificate(ctx, pgInstance.Spec.ForProvider.Zone.String())
```

apis/exoscale/v1/dbaas_types.go
```diff
@@ -115,6 +115,12 @@ func (s SizeSpec) Equals(other SizeSpec) bool {
 // A value of `0.0.0.0/0` will open the service to all addresses on the public internet.
 type IPFilter []string
 
+// +kubebuilder:validation:Enum=ch-gva-2;ch-dk-2;de-fra-1;de-muc-1;at-vie-1;bg-sof-1
+// +kubebuilder:validation:Required
```
ccremer, line 119: I would keep the Required marker in the higher-order struct
```suggestion
```

apis/exoscale/v1/postgresql_types.go
```diff
@@ -15,11 +15,8 @@ type PostgreSQLParameters struct {
 	Maintenance MaintenanceSpec `json:"maintenance,omitempty"`
 	Backup      BackupSpec      `json:"backup,omitempty"`
 
-	// +kubebuilder:validation:Enum=ch-gva-2;ch-dk-2;de-fra-1;de-muc-1;at-vie-1;bg-sof-1
-	// +kubebuilder:validation:Required
```
ccremer, line 19: I would keep the Required marker here. In case the Zone type is used in another Go project, it doesn't also become required (e.g. in the XRD code generator)

operator/webhook/validate.go
```diff
@@ -0,0 +1,28 @@
...
+	for k, v := range m {
+		switch v.(type) {
+		case string:
+		case int64:
+		case float64:
+		case bool:
+			continue
+		default:
+			return fmt.Errorf("validate: value of key %q is not a supported type (only strings, boolean and numbers): %v", k, v)
+		}
```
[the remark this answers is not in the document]
mweibel:
```
unable to create instance: Post \"https://api-ch-dk-2.exoscale.com/v2/dbaas-redis/redis-local-dev\": invalid request: Incorrect settings: Invalid input for redis_maxmemory_policy: {'test': 'ing'} is not of type 'string', 'null'"}
```

when submitting:
```
    redisSettings:
      foo:
        - bar
        - baz
      maxmemory_policy: { test: ing }
```

when removing the `maxmemory_policy` that goes through without issues and the setting is just not applied. 
I believe we should leave it at that for the moment because it seems sensible to me and revisit if we need later more. A full coverage of valid values is probably hard to do.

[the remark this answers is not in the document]
ccremer: fine by me

ZuGao (approved):
LGTM

internal/operatortest/envtest.go
```diff
@@ -92,6 +97,22 @@ func (ts *Suite) SetupSuite() {
...
+	ts.Require().NoError(err)
+	ts.Require().NotNil(mgr)
+	ts.Manager = mgr
+
+	ts.ExoClientMock = &ClientWithResponsesInterface{}
+
+	go func() {
+		err = mgr.Start(ts.Context)
+		ts.Require().NoError(err)
+	}()
```
ccremer, line 115: Why do we need a running manager in the background?

operator/rediscontroller/controller_test.go
```diff
@@ -0,0 +1,339 @@
...
+
+	ts.ExoClientMock.AssertExpectations(ts.T())
+}
+
+// reconcile waits reconcileWaitPeriod before running a reconciliation.
+func (ts *RedisControllerTestSuite) reconcile(req reconcile.Request) (reconcile.Result, error) {
+	ts.T().Helper()
+	time.Sleep(ts.reconcileWaitPeriod)
+	return ts.reconciler.Reconcile(ts.Context, req)
+}
```
ccremer, line 237: Why is the time delay needed?

operator/rediscontroller/controller_test.go
```diff
@@ -0,0 +1,339 @@
...
+					TerminationProtection: false,
+					Size: exoscalev1.SizeSpec{
+						Plan: "testplan",
+					},
+					IPFilter: exoscalev1.IPFilter{"0.0.0.0/0"},
+				},
+				RedisSettings: rs,
+			},
+		},
+	}
```
ccremer, line 332: Something that I started doing recently with Kubernetes objects. Instead of wasting tons of lines with spec and braces, I do this instead:

```suggestion
	i := &exoscalev1.Redis{}
	i.Name = name
	i.Spec.ForProvider.Maintenance.DayOfWeek = oapi.DbaasServiceMaintenanceDowMonday
	i.Spec.ForProvider.Maintenance.TimeOfDay = "12:00:00"
	...
	return i
```
it works well for structs that aren't pointers.

No hard feeling though

operator/rediscontroller/observe.go
```diff
@@ -0,0 +1,169 @@
...
+	parsed, err := url.Parse(uri)
+	if err != nil {
+		return nil, err
+	}
+	password, _ := parsed.User.Password()
+	return map[string][]byte{
+		"host":      []byte(parsed.Hostname()),
+		"port":      []byte(parsed.Port()),
+		"username":  []byte(parsed.User.Username()),
+		"password":  []byte(password),
```
ccremer, line 122: Any specific reasons why these are lowercased? They reek of upper-cased env variables :) Aren't there some commonly named variables in redis-cli or so?

mweibel: that's based on what's written in the ticket. I explicitly added REDIS_URL because that's what is usually necessary. The rest is there because of the ticket description.

operator/rediscontroller/controller_test.go
mweibel: I don't think that's an improvement TBH

internal/operatortest/envtest.go
mweibel: if not started:

```
    logger.go:130: 2022-11-08T10:26:16.618+0100	DEBUG	Cannot get managed resource	{"controller": "testredis", "request": "/test-create", "error": "the cache is not started, can not read objects"}

```

the manager starts the runnables which start the cache, I guess that's what is missing otherwise. Didn't investigate more.

operator/rediscontroller/controller_test.go
mweibel: after reconcile there's often the case that the resource status gets updated. This gives it time to persist the changes. 
If removed:

```
        	            	cannot update managed resource status: Operation cannot be fulfilled on redis.exoscale.crossplane.io "test-create": the object has been modified; please apply your changes to the latest version and try again
```

It's a bit ugly and I can imagine that it may be different depending on the machine the test is executed on. Not sure how to avoid it without spending too much time on it. 

Do you have an idea?

internal/operatortest/envtest.go
ccremer: ugh, ugly...
Can you add a comment for the reasoning? then LGTM

operator/rediscontroller/controller_test.go
ccremer: > after reconcile there's often the case that the resource status gets updated.

any idea who/what does this? the manager in the background?

> please apply your changes to the latest version and try again

that seems to be a pretty common error in Crossplane. We get these constantly in prod as well and we were unable to get rid of them yet... It's entirely possible that this is a bug in Crossplane runtime itself and not in our code.

mweibel: I guess that's just the normal [`Reconcile`](https://github.com/crossplane/crossplane-runtime/blob/master/pkg/reconciler/managed/reconciler.go#L637) which does it. Usually crossplane seems to do two or more reconciles for one change to make sure everything is done and safe. Not sure if they do too many updates or what the issue is there. But basicly every return in that function at least sets conditions, which may or may not always lead to a status update.

internal/operatortest/envtest.go
mweibel: indeed. Added a comment :+1:

operator/rediscontroller/controller_test.go
ccremer: yeah... conditions are set but the status not updated. Which indicates that a higher abstraction (manager?) does the status updates.
I don't like this, since the tests contain now delays and that makes the test unstable as it can easily fail in other environments and introduce hard-to-find bugs in test code...
Anyway, let's leave it as-is with a comment
```suggestion
// reconcile waits reconcileWaitPeriod before running a reconciliation.
func (ts *RedisControllerTestSuite) reconcile(req reconcile.Request) (reconcile.Result, error) {
	ts.T().Helper()
	// wait a bit so that status updates are done in the background (by the manager?)
	time.Sleep(ts.reconcileWaitPeriod)
	return ts.reconciler.Reconcile(ts.Context, req)
}
```

mweibel: mh the status is [always updated too](https://github.com/crossplane/crossplane-runtime/blob/master/pkg/reconciler/managed/reconciler.go#L675):
```
		return reconcile.Result{}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
```

^ that's btw the reason why I don't like the `errors.Wrap()`. It makes code harder to read ;)

Agreeing on the not liking this. I'd prefer to find the root cause and fix it. We can discuss if I should invest a day or two into that I guess? Let's talk later. 
I added the comment :+1:

ccremer (approved):

PR #61, opened 2022-11-24
Fix panic in node state mapping

glrf:
## Summary

Node states are not guaranteed to have a state. For example Kafka nodes don't. In any case we should never panic in the node state mapping

## Checklist

- [x] Update tests.


mweibel (approved):

ZuGao (approved):
LGTM

PR #71, opened 2022-12-12
Revise dbaas settings compare function

zugao:
## Summary

* Some DBaaS such as OpenSearch or MySQL are having default settings from provider. We need to make sure we don't trigger a never ending update on the provider resource.

## Checklist

- [x] I have update the integration tests

### DIFF
--- a/operator/kafkacontroller/observe_test.go
+++ b/operator/kafkacontroller/observe_test.go
@@ -337,7 +337,7 @@ func TestObserve_Outdated_Settings(t *testing.T) {
 	})
 	instance.Spec.ForProvider.KafkaSettings = setting
 	found := sampleAPIKafka("foo")
-	found.KafkaRestSettings = &map[string]interface{}{
+	found.KafkaSettings = &map[string]interface{}{
 		"foo":   "bar",
 		"count": 2,
 	}
@@ -351,7 +351,7 @@ func TestObserve_Outdated_Settings(t *testing.T) {
 		assert.NoError(t, err)
 		require.NotNil(t, res)
 		assert.True(t, res.ResourceExists, "report resource exits")
-		assert.False(t, res.ResourceUpToDate, "report resource not uptodate")
+		assert.False(t, res.ResourceUpToDate, "report resource not updated")
 	})
 }
 
--- a/operator/mapper/dbaas.go
+++ b/operator/mapper/dbaas.go
@@ -31,17 +31,18 @@ func set(s []string) map[string]struct{} {
 	return m
 }
 
-func CompareSettings(a, b runtime.RawExtension) bool {
-	sa, err := ToMap(a)
+// CompareSettings compares 2 instances of service settings
+// Some DBaaS add default settings to their services, which lead to unequal settings between
+// managed and provider's resource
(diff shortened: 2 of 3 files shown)


operator/kafkacontroller/observe_test.go
```diff
@@ -337,7 +337,7 @@ func TestObserve_Outdated_Settings(t *testing.T) {
 	})
 	instance.Spec.ForProvider.KafkaSettings = setting
 	found := sampleAPIKafka("foo")
-	found.KafkaRestSettings = &map[string]interface{}{
+	found.KafkaSettings = &map[string]interface{}{
 		"foo":   "bar",
 		"count": 2,
 	}
```
glrf, line 343: Ups thank's for catching that

operator/mapper/dbaas.go
```diff
@@ -31,19 +31,20 @@ func set(s []string) map[string]struct{} {
...
 		return false
 	}
-	sb, err := ToMap(b)
+	sb, err := ToMap(provider)
 	if err != nil {
 		return false
 	}
-	if len(sa) != len(sb) {
-		return false
-	}
```
glrf, line 46: This is technically not correct. We're now unable to remove settings.

Let's say we set a setting `bar` and later we remove that setting. This function now won't detect the change and will not revert the setting to the default

glrf (changes requested):

operator/kafkacontroller/observe_test.go
```diff
@@ -351,7 +351,7 @@ func TestObserve_Outdated_Settings(t *testing.T) {
 		assert.NoError(t, err)
 		require.NotNil(t, res)
 		assert.True(t, res.ResourceExists, "report resource exits")
-		assert.False(t, res.ResourceUpToDate, "report resource not uptodate")
+		assert.False(t, res.ResourceUpToDate, "report resource not updated")
```
susana-garcia, line 354: I wouldn't this change, the resource is actually "not up to date", right?

ZuGao: ah yes, I probably miss read it.

operator/mapper/dbaas.go
glrf: The best I could come up with for kafka rest settings was this... https://github.com/vshn/provider-exoscale/blob/master/operator/kafkacontroller/observe.go#L213-L247

Maybe you have a better idea :sweat_smile:

ZuGao: I haven't noticed that part, I was wondering why you are using Rest settings :)) As you noticed my solution was very easy and very wrong. Lets see, maybe I can come up with something tomorrow.

ZuGao: Now that I'm checking it better I don't thing it's a very good approach hardcoding the defaults that arrive from exoscale... with the next version this might change

glrf: Yeah it's not a great approach, but I don't have a better one

ZuGao: Your approach is the only one that's working therefore it is the best period.

PR #73, opened 2022-12-14
improve version validation on DBaaS 

susana-garcia:
## Summary

Improve version validation on DBaaS creation

## Checklist

- [X] Categorize the PR by setting a good title and adding one of the labels:
- [x] Update tests.
- [X] I have run `make test-e2e` and e2e tests pass successfully

### DIFF
--- a/operator/opensearchcontroller/webhook.go
+++ b/operator/opensearchcontroller/webhook.go
@@ -5,26 +5,74 @@ import (
 	"fmt"
 
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
+	exoscalesdk "github.com/exoscale/egoscale/v2"
 	"github.com/go-logr/logr"
 	exoscalev1 "github.com/vshn/provider-exoscale/apis/exoscale/v1"
 	"github.com/vshn/provider-exoscale/operator/mapper"
+	"github.com/vshn/provider-exoscale/operator/pipelineutil"
 	"github.com/vshn/provider-exoscale/operator/webhook"
 	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-var admittedVersions = []string{"1", "2"}
+const serviceType = "opensearch"
 
 // Validator validates admission requests.
 type Validator struct {
-	log logr.Logger
+	log  logr.Logger
+	kube client.Client
 }
 
 // ValidateCreate implements admission.CustomValidator.
-func (v *Validator) ValidateCreate(_ context.Context, obj runtime.Object) error {
-	openSearchInstance := obj.(*exoscalev1.OpenSearch)
+func (v *Validator) ValidateCreate(ctx context.Context, obj runtime.Object) error {
+	openSearchInstance, ok := obj.(*exoscalev1.OpenSearch)
+	if !ok {
+		return fmt.Errorf("invalid managed resource type %T for opensearch webhook", obj)
+	}
 	v.log.V(1).Info("validate create")
 
+	availableVersions, err := v.getAvailableVersions(ctx, obj)
+	if err != nil {
+		return err
+	}
+
+	err = v.validateVersion(ctx, obj, *availableVersions)
+	if err != nil {
+		return err
+	}
+
 	return v.validateSpec(openSearchInstance)
+}
+
+func (v *Validator) getAvailableVersions(ctx context.Context, obj runtime.Object) (*[]string, error) {
+	openSearchInstance := obj.(*exoscalev1.OpenSearch)
+
+	v.log.V(1).Info("get opensearch available versions")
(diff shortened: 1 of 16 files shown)


wejdross:
I didn't test it locally, but it looks good for me.

operator/opensearchcontroller/webhook.go
```diff
@@ -114,7 +112,18 @@ func (v *Validator) compareZone(old, new *exoscalev1.OpenSearch) error {
 }
 
 func (v *Validator) compareVersion(old, new *exoscalev1.OpenSearch) error {
```
zugao, line 114: I believe we should not allow to change the version therefore I would make this field immutable. Changing the version would imply that the customer would wrongly believe that he can change the version of his running instance. @Kidswiss do you agree with that?
P.S: Applies to all DBaaS

zugao (commented):
You should update the title of this PR as it is misleading. You also updated the Update process of a DBaaS not just Create.

I have to admit that I don't really like where this whole version management is going. Hard coding the version in webhooks feels very wrong. I think we are doing the same mistake Exoscale did, exposing the version which ultimately you cannot choose. At this point I would propose to remove version field from all DBaaS and update our User Docs to let them know which single version we support thus manually maintaining the versions of Exoscale DBaaS. Yes there are services where you can choose an older version such as OpenSearch but unless Exoscale makes it more clear via their API which version is available instead of us "guessing" I would leave opt out on exposing this field. Let's see what other members think about this.

susana-garcia:
@zugao I wasn't sure about this approach, that's why I asked yesterday in rocket.chat, before making the changes in all DBaaS...

Kidswiss:
> Yes there are services where you can choose an older version such as OpenSearch but unless Exoscale makes it more clear via their API which version is available instead of us "guessing" I would opt out on exposing this field.

I just found out that they have a CLI that can show you the versions they provide. This eliminates the guessing. It's not just OpenSearch where you have multiple available, but PostgreSQL as well:

```bash
exo dbaas type show pg
┼────────────────────┼───────────────────────────────────────────────────────────┼
│ Name               │ pg                                                        │
│ Description        │ PostgreSQL - Object-Relational Database Management System │
│ Available Versions │ 11, 12, 13, 14                                            │
│ Default Version    │ 14                                                        │
┼────────────────────┼───────────────────────────────────────────────────────────┼

➜ exo dbaas type show mysql
┼────────────────────┼───────────────────────────────────────────────┼
│ Name               │ mysql                                         │
│ Description        │ MySQL - Relational Database Management System │
│ Available Versions │ 8                                             │
│ Default Version    │ 8                                             │
┼────────────────────┼───────────────────────────────────────────────┼

➜ exo dbaas type show redis
┼────────────────────┼────────────────────────────────────────┼
│ Name               │ redis                                  │
│ Description        │ Redis - In-Memory Data Structure Store │
│ Available Versions │                                        │
│ Default Version    │ -                                      │
┼────────────────────┼────────────────────────────────────────┼

➜ exo dbaas type show kafka
┼────────────────────┼──────────────────────────────────────────────────────┼
│ Name               │ kafka                                                │
│ Description        │ Kafka - High-Throughput Distributed Messaging System │
│ Available Versions │ 3.2                                                  │
│ Default Version    │ 3.2                                                  │
┼────────────────────┼──────────────────────────────────────────────────────┼

➜ exo dbaas type show opensearch
┼────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┼
│ Name               │ opensearch                                                                          │
│ Description        │ OpenSearch - Search & Analyze Data in Real Time, derived from Elasticsearch v7.10.2 │
│ Available Versions │ 1, 2                                                                                │
│ Default Version    │ 2                                                                                   │
┼────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┼
```

So my proposal: do it for the services where there are actually multiple versions available. If they add new versions in the future for mysql or kafka, then we add it there as well.

glrf:
In my opinion at a provider level we should expose versions the same way as exoscale does, even if currently only one version is valid. I agree though that hard coding this isn't the best solution. 

Given that their CLI exposes the available versions maybe there is a hidden API that exposes this?

zugao:
> Given that their CLI exposes the available versions maybe there is a hidden API that exposes this?

We should open a ticket to exoscale as this endpoint does not exist in the open API though I doubt they have a hidden endpoint. They probably hard coded those versions themselves.

susana-garcia:
Just found a type in the oapi `DbaasServiceType` that has `AvailableVersions` and you can get a list `ListDbaasServiceTypes`. I will use that instead of magic strings.
Thank you for the inputs.

zugao:
Since now we enforce version field for kafka we need to also add a [default value](https://github.com/vshn/component-appcat/blob/master/packages/composite/dbaas/exoscale/kafka.yml#L87) in the component.

glrf (approved):

PR #82, opened 2023-01-23
bugfix: return error when iam key or secret are not present

susana-garcia:
bugfix: return an error when iam key or secret are not present when setting connection details

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/operator/iamkeycontroller/pipeline.go
+++ b/operator/iamkeycontroller/pipeline.go
@@ -2,6 +2,8 @@ package iamkeycontroller
 
 import (
 	"context"
+	"errors"
+
 	exoscalesdk "github.com/exoscale/egoscale/v2"
 	exoscalev1 "github.com/vshn/provider-exoscale/apis/exoscale/v1"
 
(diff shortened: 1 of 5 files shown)


operator/iamkeycontroller/pipeline.go
```diff
@@ -61,11 +63,17 @@ func NewPipeline(client client.Client, recorder event.Recorder, exoscaleClient *
 	}
 }
 
-func toConnectionDetails(iamKey *exoscalesdk.IAMAccessKey) managed.ConnectionDetails {
+func toConnectionDetails(iamKey *exoscalesdk.IAMAccessKey) (managed.ConnectionDetails, error) {
+	if iamKey.Key == nil {
+		return nil, fmt.Errorf("iamKey key not found in connection details")
```
mweibel, line 68:
```suggestion
		return nil, errors.New("iamKey key not found in connection details")
```

`fmt.Errorf` is used when you need a format string in the error which gets replaced by something (usually `%w` which wraps an error). If you just need a normal error value, `errors.New()` is the way to go.

operator/iamkeycontroller/pipeline.go
```diff
@@ -61,11 +63,17 @@ func NewPipeline(client client.Client, recorder event.Recorder, exoscaleClient *
...
+	}
+	if iamKey.Secret == nil {
+		return nil, fmt.Errorf("iamKey secret not found in connection details")
```
mweibel, line 71:
```suggestion
		return nil, errors.New("iamKey secret not found in connection details")
```

mweibel (commented):
slightly nit comment but might be good to fix. Otherwise looks good :+1:

operator/iamkeycontroller/pipeline.go
susana-garcia: true that, cheers

mweibel (approved):

glrf (approved):
LGTM, should fix the panic

PR #84, opened 2023-01-30
bugfix: do not access connection details if credentials secret is not found

susana-garcia:
## Summary

bugfix: do not access connection details if credentials secret is not found in iam key observer

## Checklist

      changelog.
      linked in the changelog.

### DIFF
--- a/operator/iamkeycontroller/observe.go
+++ b/operator/iamkeycontroller/observe.go
@@ -42,18 +42,15 @@ func (p *IAMKeyPipeline) Observe(ctx context.Context, mg resource.Managed) (mana
 	}
 
 	pipe := pipeline.NewPipeline[*pipelineContext]()
-	result := pipe.WithBeforeHooks(pipelineutil.DebugLogger(pctx)).
+	err = pipe.WithBeforeHooks(pipelineutil.DebugLogger(pctx)).
 		WithSteps(
 			pipe.NewStep("fetch credentials secret", p.fetchCredentialsSecret),
 			pipe.NewStep("check credentials", p.checkSecret),
 		).RunWithContext(pctx)
 
-	if result != nil {
-		connDetails, err := toConnectionDetails(pctx.iamExoscaleKey)
-		if err != nil {
-			return managed.ExternalObservation{}, fmt.Errorf("cannot parse connection details: %w", err)
-		}
-		return managed.ExternalObservation{ResourceExists: true, ResourceUpToDate: false, ConnectionDetails: connDetails}, nil
+	if err != nil {
+		log.V(2).Info("pipeline that fetches and checks credentials secret returned", "error", err)
+		return managed.ExternalObservation{ResourceExists: true, ResourceUpToDate: false}, nil
 	}
 
 	iamKey.Status.AtProvider.KeyName = *pctx.iamExoscaleKey.Name


operator/iamkeycontroller/observe.go
```diff
@@ -49,11 +49,9 @@ func (p *IAMKeyPipeline) Observe(ctx context.Context, mg resource.Managed) (mana
 		).RunWithContext(pctx)
 
 	if result != nil {
-		connDetails, err := toConnectionDetails(pctx.iamExoscaleKey)
```
zugao, line 52: You skip to connection details altogether, isn't it necessary in some cases?

operator/iamkeycontroller/observe.go
```diff
@@ -49,11 +49,9 @@ func (p *IAMKeyPipeline) Observe(ctx context.Context, mg resource.Managed) (mana
...
-		if err != nil {
-			return managed.ExternalObservation{}, fmt.Errorf("cannot parse connection details: %w", err)
-		}
-		return managed.ExternalObservation{ResourceExists: true, ResourceUpToDate: false, ConnectionDetails: connDetails}, nil
+		// it may happen that the resource, which secret the observe is trying to fetch, was just deleted.
```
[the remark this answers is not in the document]
glrf: I agree, a bit more in detail comment might help.

glrf (approved):
LGTM

operator/iamkeycontroller/observe.go
susana-garcia: The problem is that the pipeline fetches the credentials secret and checks the secret, if that fails (because e.g. the IAM key was just deleted), it doesn't make sense to try to fetch the connection details, as they might be already gone too.

operator/iamkeycontroller/observe.go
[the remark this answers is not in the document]
susana-garcia: Yes, that's true. I think I might just delete it completely. In other observers we don't fetch the connection details unless the previous steps were successful, so we could say that's the default behavior, what do you think?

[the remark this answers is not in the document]
glrf: Maybe you're right. The comment confuses more that it helps. (What also IMO doesn't help is that it's `result` and not `err` but that's not part of this PR..)

operator/iamkeycontroller/observe.go
zugao: Then I'm ok with it.

operator/iamkeycontroller/observe.go
```diff
@@ -49,11 +49,8 @@ func (p *IAMKeyPipeline) Observe(ctx context.Context, mg resource.Managed) (mana
 		).RunWithContext(pctx)
 
 	if result != nil {
-		connDetails, err := toConnectionDetails(pctx.iamExoscaleKey)
-		if err != nil {
-			return managed.ExternalObservation{}, fmt.Errorf("cannot parse connection details: %w", err)
-		}
-		return managed.ExternalObservation{ResourceExists: true, ResourceUpToDate: false, ConnectionDetails: connDetails}, nil
+		log.V(2).Info("pipeline that fetches and checks credentials secret returned", "error", err)
```
glrf, line 52: I think we actually need `result` not `err`

```suggestion
		log.V(2).Info("pipeline that fetches and checks credentials secret returned", "error", result)
```

glrf (changes requested):

operator/iamkeycontroller/observe.go
susana-garcia: sorry, yes, saw it later. Actually it makes more sense to call it `err`

glrf (approved):

zugao (approved):
LGTM

PR #85, opened 2023-01-31
bypass governance retention when deleting s3 bucket

susana-garcia:
## Summary 

s3 seems to use the default retention governance mode, which causes bucket deletion to fail unless we bypass the governance retention.

https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock-overview.html

## Checklist

      changelog.
      linked in the changelog.
- [X] Link this PR to related issues or PRs.

### DIFF
--- a/operator/bucketcontroller/delete.go
+++ b/operator/bucketcontroller/delete.go
@@ -3,6 +3,7 @@ package bucketcontroller
 import (
 	"context"
 	"fmt"
+
 	pipeline "github.com/ccremer/go-command-pipeline"
 	"github.com/crossplane/crossplane-runtime/pkg/errors"
 	"github.com/crossplane/crossplane-runtime/pkg/event"
@@ -55,7 +56,9 @@ func (p *ProvisioningPipeline) deleteAllObjects(ctx *pipelineContext) error {
 		}
 	}()
 
-	for obj := range p.minioClient.RemoveObjects(ctx, bucketName, objectsCh, minio.RemoveObjectsOptions{GovernanceBypass: true}) {
+	// s3 seems to use the default retention governance mode, which causes bucket deletion to fail, unless bypassing governance retention
+	// https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock-overview.html
+	for obj := range p.minioClient.RemoveObjects(ctx, bucketName, objectsCh, minio.RemoveObjectsOptions{GovernanceBypass: false}) {
 		return fmt.Errorf("object %q cannot be removed: %w", obj.ObjectName, obj.Err)
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


Kidswiss:
It looks like Exoscale supports this feature, as it's not on their unsupported list: https://community.exoscale.com/documentation/storage/quick-start/#s3-unsupported-features

So I added a bit more logic to determine if ObjectLock is enabled or not on a given bucket. If it's enabled we need to bypass it. Otherwise we would run into errors if a customer enables it on a bucket they provisioned via the provider.

@glrf @zugao : we should probably port this to provider-cloudscale as well?

operator/bucketcontroller/delete.go
```diff
@@ -55,12 +56,28 @@ func (p *ProvisioningPipeline) deleteAllObjects(ctx *pipelineContext) error {
 		}
 	}()
 
-	for obj := range p.minioClient.RemoveObjects(ctx, bucketName, objectsCh, minio.RemoveObjectsOptions{GovernanceBypass: true}) {
+	bypassGovernance, err := p.isBucketLockEnabled(ctx, bucketName)
+	if err != nil {
+		return fmt.Errorf("cannot determine bucketLockConfig of bucket %s: %w", bucketName, err)
+	}
+
+	for obj := range p.minioClient.RemoveObjects(ctx, bucketName, objectsCh, minio.RemoveObjectsOptions{GovernanceBypass: bypassGovernance}) {
```
glrf, line 64: Not sure if I understand the bug this fixes. So if we specify to bypass governance, but bucket lock is not enabled, we get an error from exoscale, or why can't we unconditionally bypass it?

Kidswiss: Exactly, if it's not enabled and we set it to `true` it will throw an error and vice-versa.

glrf: O wow, I'd expect them to ignore the bypass field and delete the objects if we specify it for a bucket without locks..

operator/bucketcontroller/delete.go
```diff
@@ -55,12 +56,28 @@ func (p *ProvisioningPipeline) deleteAllObjects(ctx *pipelineContext) error {
 		}
 	}()
 
-	for obj := range p.minioClient.RemoveObjects(ctx, bucketName, objectsCh, minio.RemoveObjectsOptions{GovernanceBypass: true}) {
+	bypassGovernance, err := p.isBucketLockEnabled(ctx, bucketName)
+	if err != nil {
+		return fmt.Errorf("cannot determine bucketLockConfig of bucket %s: %w", bucketName, err)
```
glrf, line 61: Nitpick: I'd probably suggest that in case of an error, we log the error, assume that there is no lock and try to delete the objects anyway, instead of aborting.

Mainly because of the brittle error matching below, but also if there maybe are other errors that would still allow us to delete the objects.

glrf (approved):

operator/bucketcontroller/delete.go
Kidswiss: Fine by me, if the deletion of the objects fails we'll still see it as an event then.

PR #92, opened 2023-03-29
Remove SOS IAMkey access tag

glrf:
Setting the sos tag will ignore any set oprations and will allow all SOS operations, which is not what we want.

## Checklist

      changelog.
      linked in the changelog.


operator/iamkeycontroller/create.go
```diff
@@ -74,7 +74,6 @@ func (p *IAMKeyPipeline) createIAMKey(ctx *pipelineContext) error {
 		// Some permissions are excluded such as create and list all sos buckets.
 		exoscalesdk.CreateIAMAccessKeyWithOperations(IAMKeyAllowedOperations),
 		// The tag and resource are used to limit the permissions to object storage
```
zugao, line 76: Remove this line as well, it is not needed anymore :)

glrf: A fair, thanks. Half of the comment is wrong now :sweat_smile:

PR #96, opened 2023-06-28
Migrating from /access-key to iam-role/ and iam-key/ endpoints to generate iam keys

wejdross:
After Exoscale forced us to move to new iam-role and iam-keys endpoints in their API provider-exoscale wasn't able to issue buckets anymore. 

this merge request fix that

## Checklist

      changelog.
      linked in the changelog.
- [ ] I have run successfully `make test-e2e` locally.
- [ ] Update the documentation.
- [ ] Link this PR to related issues or PRs.


operator/iamkeycontroller/create.go
```diff
@@ -54,45 +61,157 @@ func (p *IAMKeyPipeline) Create(ctx context.Context, mg resource.Managed) (manag
...
+						"sos": {
+							Type: &policyRules,
+							Rules: &[]exooapi.IamServicePolicyRule{
+								{
+									Action: &policyDeny,
+									// bucket name in this rule must be in single quotes, otherwise it doesn't work and user has access to everything
+									Expression: pointer.String("resources.bucket != " + "'" + strings.Split(iamKey.Spec.ForProvider.KeyName, ".")[1] + "'"),
+								},
+								{
+									Action:     &policyDeny,
```
zugao, line 121: I read the support ticket you opened and it mentioned that it should be `policy allow = true`. I guess this should work as well?

operator/iamkeycontroller/observe.go
```diff
@@ -53,68 +55,68 @@ func (p *IAMKeyPipeline) Observe(ctx context.Context, mg resource.Managed) (mana
 		return managed.ExternalObservation{ResourceExists: true, ResourceUpToDate: false}, nil
 	}
 
+	log.V(1).Info("Observing resource line 58")
 	iamKey.Status.AtProvider.KeyName = *pctx.iamExoscaleKey.Name
-	iamKey.Status.AtProvider.ServicesSpec.SOS.Buckets = getBuckets(*pctx.iamExoscaleKey.Resources)
```
zugao, line 57: Now as I understand it's only one bucket from the role?

zugao (commented):
Haven't finished the review.

operator/iamkeycontroller/create.go
```diff
@@ -54,45 +61,157 @@ func (p *IAMKeyPipeline) Create(ctx context.Context, mg resource.Managed) (manag
...
+		log.Info("IAM Key doesnt exists, creating", "keyName", ctx.iamKey.Spec.ForProvider.KeyName)
+		newIamKey := exooapi.IamApiKey{
+			RoleId: &iamRoleID,
+			Name:   &ctx.iamKey.Spec.ForProvider.KeyName,
+		}
+
+		_, err = ExecuteRequest(ctx, "GET", ctx.iamKey.Spec.ForProvider.Zone, "/v2/iam-role/"+iamRoleID, p.apiKey, p.apiSecret, nil)
+		if err != nil {
+			log.Error(err, "IAM Role doesn't exist (yet), sleeping 3 seconds before creating a key", "iamrole", iamRoleID)
+			time.Sleep(3 * time.Second)
```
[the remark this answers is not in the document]
wejdross: that's just leftover, deleted

operator/iamkeycontroller/observe.go
```diff
@@ -53,68 +55,68 @@ func (p *IAMKeyPipeline) Observe(ctx context.Context, mg resource.Managed) (mana
 		return managed.ExternalObservation{ResourceExists: true, ResourceUpToDate: false}, nil
 	}
 
+	log.V(1).Info("Observing resource line 58")
```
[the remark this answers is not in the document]
wejdross: yeah sure, it was debugging statement

operator/iamkeycontroller/delete.go
```diff
@@ -33,15 +33,24 @@ func (p *IAMKeyPipeline) deleteIAMKey(ctx *pipelineContext) error {
 	log := controllerruntime.LoggerFrom(ctx)
 	iamKey := ctx.iamKey
 
-	err := p.exoscaleClient.RevokeIAMAccessKey(ctx, iamKey.Spec.ForProvider.Zone, &exoscalesdk.IAMAccessKey{
-		Key: &iamKey.Status.AtProvider.KeyID,
-	})
+	// we know
```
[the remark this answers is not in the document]
wejdross: i don't know what was that, deleted

operator/iamkeycontroller/create.go
wejdross: fixed

operator/iamkeycontroller/observe.go
zugao: We need to update that in the status field. Also you will have to distinguish between old keys and new keys...

operator/iamkeycontroller/create.go
```diff
@@ -54,45 +61,157 @@ func (p *IAMKeyPipeline) Create(ctx context.Context, mg resource.Managed) (manag
...
 	}
 
-	exoscaleIAM, err := p.exoscaleClient.CreateIAMAccessKey(ctx, zone, iamKeyName, iamKeyOpts...)
+	err = json.NewDecoder(resp.Body).Decode(&keyList)
 	if err != nil {
+		log.Error(err, "Cannot unmarshall response to keyList")
 		return err
 	}
 
-	// Limitation by crossplane: The interface managed.ExternalClient doesn't allow updating the resource during creation except annotations.
```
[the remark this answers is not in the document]
zugao: I think I remembered this issue. I would rephrase a bit the sentence:
`// Limitation by crossplane: The interface managed.ExternalClient doesn't allow updating  resource in creation reconcile with the except of annotations`

operator/iamkeycontroller/observe.go
wejdross: fixed in newest commit

operator/bucketcontroller/observe.go
```diff
@@ -14,6 +14,7 @@ import (
 )
 
 var bucketExistsFn = func(ctx context.Context, mc *minio.Client, bucketName string) (bool, error) {
+	//time.Sleep(time.Second * 5)
```
zugao, line 17: This should be removed

zugao (changes requested):
I was not able to create a new iamkey... We need tests as well for this if we don't want surprises.

operator/iamkeycontroller/delete.go
```diff
@@ -33,15 +39,36 @@ func (p *IAMKeyPipeline) deleteIAMKey(ctx *pipelineContext) error {
...
-	})
-	if err != nil {
-		return err
+	log.Info("Starting IAM key deletion", "keyName", iamKey.Spec.ForProvider.KeyName)
+
+	fmt.Println(iamKey.ObjectMeta.Annotations)
+
+	if _, ok := iamKey.ObjectMeta.Annotations[RoleIDAnnotationKey]; ok {
+
+		_, err := ExecuteRequest(ctx, "DELETE", ctx.iamKey.Spec.ForProvider.Zone, "/v2/api-key/"+iamKey.Status.AtProvider.KeyID, p.apiKey, p.apiSecret, nil)
```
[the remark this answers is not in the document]
wejdross: yeah, that happened because:
1) exoscale creates a role and returns us ID
2) You can immediately GET this role by ID, but...
3) You couldn't create resource referencing to this role-id immediately (they have probably some sync on datacenters)

I added small for loop which executes query and blocks for 500ms after fail as I find it cleaner solution than regular time.Sleep(time.Second * 10)

Manual tests shows issue is solved

operator/iamkeycontroller/create.go
```diff
@@ -54,45 +63,121 @@ func (p *IAMKeyPipeline) Create(ctx context.Context, mg resource.Managed) (manag
...
+				DefaultServiceStrategy: exooapi.IamPolicyDefaultServiceStrategyDeny,
+				Services: exooapi.IamPolicy_Services{
+					AdditionalProperties: map[string]exooapi.IamServicePolicy{
+						"sos": {
+							Type: &policyRules,
+							Rules: &[]exooapi.IamServicePolicyRule{
+								{
+									Action: &policyDeny,
+									// bucket name in this rule must be in single quotes, otherwise it doesn't work and user has access to everything
+									Expression: pointer.String("resources.bucket != " + "'" + strings.Split(iamKey.Spec.ForProvider.KeyName, ".")[1] + "'"),
```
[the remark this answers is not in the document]
wejdross: caused by issue described below

operator/iamkeycontroller/create.go
```diff
@@ -54,45 +63,121 @@ func (p *IAMKeyPipeline) Create(ctx context.Context, mg resource.Managed) (manag
...
+		return err
+	}
 
-	var keyResources []exoscalesdk.IAMAccessKeyResource
-	for _, bucket := range iamKey.Spec.ForProvider.Services.SOS.Buckets {
-		keyResource := exoscalesdk.IAMAccessKeyResource{
-			Domain:       SOSResourceDomain,
-			ResourceName: bucket,
-			ResourceType: BucketResourceType,
+	for _, val := range iamRoles.IamRoles {
```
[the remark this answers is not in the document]
wejdross: because we do something realy non-crossplane here, I should rather create iamrolecontroleer and create logic for CRUD of iam-role, then add it to composition, so there would be normal Create() and observer() process, since I create 2 things in one call, they depend on each other, by this I tried to ensure that role always exists which was bad approach, it should be fixed now by retrying of IAM-KEY creation

operator/iamkeycontroller/create.go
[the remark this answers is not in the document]
zugao: It is caused by this `strings.Split(iamKey.Spec.ForProvider.KeyName, ".")[1]`.  THis approach works only when compositions are used. If you want to create a simple iamkey the whole thing falls apart.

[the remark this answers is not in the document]
zugao: By the way why are you taking the bucket name from key name? The buckets are under `iamKey.Spec.ForProvider.Services.SOS`.

operator/iamkeycontroller/create.go
```diff
@@ -54,45 +64,110 @@ func (p *IAMKeyPipeline) Create(ctx context.Context, mg resource.Managed) (manag
...
+	iamRoleID := *iamRole.Reference.Id
+	log.Info("IAM Role created", "iamRoleID", iamRoleID)
+
+	log.Info("IAM Key doesnt exists, creating", "keyName", ctx.iamKey.Spec.ForProvider.KeyName)
+	newIamKey := exooapi.IamApiKey{
+		RoleId: &iamRoleID,
+		Name:   &ctx.iamKey.Spec.ForProvider.KeyName,
+	}
+
+	b1, _ := json.MarshalIndent(newIamKey, "\t", "\t")
```
zugao, line 131: Please check the code before sending for a review. There's a lost of debug code everywhere

operator/iamkeycontroller/create.go
```diff
@@ -54,45 +64,110 @@ func (p *IAMKeyPipeline) Create(ctx context.Context, mg resource.Managed) (manag
...
-	for _, bucket := range iamKey.Spec.ForProvider.Services.SOS.Buckets {
-		keyResource := exoscalesdk.IAMAccessKeyResource{
-			Domain:       SOSResourceDomain,
-			ResourceName: bucket,
-			ResourceType: BucketResourceType,
+	b1, _ = json.MarshalIndent(iamRole, "\t", "\t")
+	fmt.Println("\n\n\n\n\n\n\n\n\n\n", string(b1))
+
+	// since their API is async and it needs a moment to create the IAM Role, we need to wait for it
+	for i := 0; i < 10; i++ {
```
zugao, line 138: Nitpick: This is ugly. We should implement a [RoundTripper](https://morioh.com/p/5beb43562e87) for the client and abstract this loop

operator/iamkeycontroller/create.go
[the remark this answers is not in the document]
wejdross: If I create separate role, then detection of legacy key and new key would be terribly hard, with current implementation, my goal is to ensure observability and deletion of old keys and CREATE of new types of key, so with time we will simply  replace old keys with newer ones and we can do the refactor

operator/iamkeycontroller/delete.go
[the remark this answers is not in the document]
wejdross: That should be fixed right now

operator/iamkeycontroller/create.go
wejdross: Ok, I've familiarized with RoundTripper and I think it's great tool but for bigger use cases, it does basically the same thing but in different place. I think that implementation of round tripper here, would just make code more complex. If You insist then no problem, but I can't see value for single request

operator/iamkeycontroller/create.go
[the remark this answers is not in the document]
wejdross: changed to that

zugao (approved):
Code looks ok now, I haven't tested the new commit though. Also make sure the new version works with the legacy keys in a cluster.

Makefile.vars.mk
```diff
@@ -18,7 +18,7 @@ CONTAINER_REGISTRY ?= ghcr.io
 UPBOUND_CONTAINER_REGISTRY ?= xpkg.upbound.io
 
 # Image URL to use all building/pushing image targets
-CONTAINER_IMG ?= $(CONTAINER_REGISTRY)/$(PROJECT_OWNER)/$(PROJECT_NAME)/controller:$(IMG_TAG)
+CONTAINER_IMG ?= ghcr.io/vshn/provider-exoscale/provider:v0.9.2
```
Kidswiss, line 21: Shouldn't we switch this back? I feel like some CI/CD will break otherwise.

Makefile.vars.mk
```diff
@@ -29,7 +29,7 @@ UPBOUND_PACKAGE_IMG ?= $(UPBOUND_CONTAINER_REGISTRY)/$(PROJECT_OWNER)/$(PROJECT_
 KIND_NODE_VERSION ?= v1.24.0
 KIND_IMAGE ?= docker.io/kindest/node:$(KIND_NODE_VERSION)
 KIND_KUBECONFIG ?= $(kind_dir)/kind-kubeconfig-$(KIND_NODE_VERSION)
-KIND_CLUSTER ?= $(PROJECT_NAME)-$(KIND_NODE_VERSION)
+KIND_CLUSTER ?= kindev
```
Kidswiss, line 32: Same here

operator/iamkeycontroller/create.go
Kidswiss: Agreed, if they start to switch everything to async some abstraction for this would be nice... (if they don't provide it in egoscale)

.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
+{
```
Kidswiss, line 1: Should we really keep this in the repo? Not all of schedar uses vscode.

operator/iamkeycontroller/observe.go
```diff
@@ -29,20 +31,34 @@ func (p *IAMKeyPipeline) Observe(ctx context.Context, mg resource.Managed) (mana
...
 		}
 	}
 
 	pctx := &pipelineContext{Context: ctx, iamKey: iamKey}
-	err := p.getIAMKey(pctx)
-	if err != nil {
-		return managed.ExternalObservation{}, resource.Ignore(isNotFound, err)
+
+	// since new create() adds RoleID annotation it's enought to just check whether it exists
+	if _, ok := iamKey.ObjectMeta.Annotations[RoleIDAnnotationKey]; ok {
```
Kidswiss, line 45: Shouldn't we delete this annotation and write it to the status? Like we do for `KeyIDAnnotationKey`.

See line 31 for how that annotation is handled.

Kidswiss (dismissed):

Kidswiss:
Also there's a typo in the PR title :D `Migratic`

.vscode/launch.json
zugao: Ah yes, we have to remove that.

operator/iamkeycontroller/observe.go
wejdross: yeah, migrated into that logic

wejdross:
Please review last small change in last commit, it fixes observation of legacy keys. Tested in local kind cluster and lab cluster. Works ^^

Makefile.vars.mk
```diff
@@ -28,8 +28,8 @@ UPBOUND_PACKAGE_IMG ?= $(UPBOUND_CONTAINER_REGISTRY)/$(PROJECT_OWNER)/$(PROJECT_
 # https://hub.docker.com/r/kindest/node/tags
 KIND_NODE_VERSION ?= v1.24.0
 KIND_IMAGE ?= docker.io/kindest/node:$(KIND_NODE_VERSION)
-KIND_KUBECONFIG ?= $(kind_dir)/kind-kubeconfig-$(KIND_NODE_VERSION)
-KIND_CLUSTER ?= $(PROJECT_NAME)-$(KIND_NODE_VERSION)
+KIND_KUBECONFIG ?= /home/wejdross/vshn/shedar/kindev/.kind/kind-kubeconfig-v1.24.4
+KIND_CLUSTER ?= kindev
```
zugao, line 32: we should not name the kind cluster the same as kinddev.

Makefile.vars.mk
```diff
@@ -28,8 +28,8 @@ UPBOUND_PACKAGE_IMG ?= $(UPBOUND_CONTAINER_REGISTRY)/$(PROJECT_OWNER)/$(PROJECT_
 # https://hub.docker.com/r/kindest/node/tags
 KIND_NODE_VERSION ?= v1.24.0
 KIND_IMAGE ?= docker.io/kindest/node:$(KIND_NODE_VERSION)
-KIND_KUBECONFIG ?= $(kind_dir)/kind-kubeconfig-$(KIND_NODE_VERSION)
-KIND_CLUSTER ?= $(PROJECT_NAME)-$(KIND_NODE_VERSION)
+KIND_KUBECONFIG ?= /home/wejdross/vshn/shedar/kindev/.kind/kind-kubeconfig-v1.24.4
```
zugao, line 31: this is wrong

Makefile.vars.mk
zugao, line 21: This will definitely break smth else.

zugao (changes requested):
Some small things

Makefile.vars.mk
wejdross: restored those files, merging

Makefile.vars.mk
wejdross: restored those files, merging

wejdross:
@zugao @Kidswiss I need at least one aprooval :P

zugao (approved):
LGTM

newbenji:
Now just wish new release with this :-D

wejdross:
Hello @newbenji, release will be done in less than an hour I guess

PR #97, opened 2023-07-06
fixing e2e tests 

wejdross:
Adding logic to IamKey controller, so  e2e tests pass. This must be handled like to, because we need to support legacy keys as well


TheBigLee (approved):
LGTM

PR #101, opened 2023-09-28
Deny bucket listing in iam policy

Kidswiss:
Previously it was possible that an IAM key could list all buckets in the project, even the ones where the IAM key didn't have any other permissions.

We now explicitly deny the listing of buckets. Also, already existing IAM keys will be updated with the new policy.

## Checklist

      changelog.
      linked in the changelog.
- [x] Link this PR to related issues or PRs.


test/e2e/kafka/00-assert.yaml
```diff
@@ -21,7 +21,7 @@ spec:
     size:
       plan: startup-2
     zone: ch-dk-2
-    version: '3.2'
+    version: '3.5'
```
wejdross, line 24: is it anyhow related?

wejdross (approved):
LGTM

operator/iamkeycontroller/update.go
```diff
@@ -5,13 +5,22 @@ import (
...
 func (p *IAMKeyPipeline) Update(ctx context.Context, mg resource.Managed) (managed.ExternalUpdate, error) {
 	log := controllerruntime.LoggerFrom(ctx)
-	log.V(1).Info("Updating resource (noop)")
-	return managed.ExternalUpdate{}, nil
+	log.V(1).Info("Updating role")
+
+	iamKey := fromManaged(mg)
+	iamKey.SetConditions(exov1.Updating())
+
+	role := createRole(iamKey.Spec.ForProvider.KeyName, iamKey.Spec.ForProvider.Services.SOS.Buckets)
```
zugao, line 21: Why do you create the role in `Update()` method?

zugao (approved):
LGTM, I haven't tested it. You might want to test it throughly as I don't see any unit/e2e tests.

Kidswiss:
> LGTM, I haven't tested it. You might want to test it throughly as I don't see any unit/e2e tests.

I've tested it via kind.

I agree about the unit/e2e test, but on the other hand this logic should actually be part of a composition instead of the provider, as it's very specific to our use-case.

operator/iamkeycontroller/update.go
Kidswiss: `createRole()` creates the role object with the policy, it doesn't create anything on the API.

Should I rename this function?

test/e2e/kafka/00-assert.yaml
Kidswiss: Somewhat, the e2e-test doesn't work anymore with 3.2.

PR #106, opened 2024-05-27
Fix version and update dependencies

zugao:
* Updates dependencies
* Updates code to work with new dependencies
* Update webhook debugging
* Disable kafka version checking when missing. Will get the latest version from Aiven.

## Checklist

      changelog.
      linked in the changelog.


README.md
```diff
@@ -39,27 +39,38 @@ See all targets with `make help`
 ### QuickStart Demonstration
 
 1. Get an API token exoscale.com
-1. `export EXOSCALE_API_KEY=<the-key>`
-1. `export EXOSCALE_API_SECRET=<the-secret>`
-1. `make local-install`
+2. `export EXOSCALE_API_KEY=<the-key>`
```
Kidswiss, line 42: No need to manually increment here. Most markdown renderers will do that for you.

operator/iamkeycontroller/create.go
```diff
@@ -162,7 +162,7 @@ func (p *IAMKeyPipeline) createCredentialsSecret(ctx *pipelineContext) error {
 		for k, v := range connDetails {
 			secret.Data[k] = v
 		}
-		secret.Immutable = pointer.Bool(true)
+		secret.Immutable = ptr.To[bool](true)
```
Kidswiss, line 165:
```suggestion
		secret.Immutable = ptr.To(true)
```

Generics can infer the type, no need to specify it.

operator/iamkeycontroller/util.go
```diff
@@ -16,11 +16,11 @@ func createRole(keyName string, buckets []string) *exooapi.IamRole {
 
 	iamRole := &exooapi.IamRole{
 		Name:        &keyName,
-		Description: pointer.String("IAM Role for SOS+IAM creation, it was autogenerated by provider-exoscale"),
+		Description: ptr.To[string]("IAM Role for SOS+IAM creation, it was autogenerated by provider-exoscale"),
```
Kidswiss, line 19: Same here.

package/package.mk
```diff
@@ -13,11 +13,11 @@ $(crossplane_bin): | $(go_bin)
 
 # Install up plugin
 $(up_bin):export GOBIN = $(go_bin)
-$(up_bin):export VERSION=v0.15.0
+$(up_bin):export VERSION=v0.30.0
```
Kidswiss, line 16: Is the up binary still necessary? Afaik the crossplane crank binary can do all the operations as well. At least we only use crank in the main appcat repo.

In fact we use these operations already here, it would be nice if we could cut down the dependencies a bit.

Also, instead of installing up/crank like we do now we could do it the same way we do in the main appcat repository:

Import it to a `tools.go` file and then use `go run github.com/crossplane/crossplane/cmd/crank`. This has the advantage that the versions of the tools are automatically managed by `go.mod`.

Kidswiss (changes requested):

README.md
zugao: Well as you can see the previous markdown didn't work quite well lol :D

package/package.mk
zugao: Thanks for the tip. I removed up all together and use crossplane binary.

README.md
Kidswiss: The current readme shows fine for me though?

https://github.com/vshn/provider-exoscale/blob/master/README.md#quickstart-demonstration

Be aware that the source code will not automatically increment the numbering, it's only in the rendered version.

But anyway it's just a nitpick.

Kidswiss (approved):

PR #122, opened 2025-01-23
Refactor provider for new Exoscale API v3

TheBigLee:
## Checklist

      changelog.
      linked in the changelog.
- [ ] Update the documentation.
- [ ] Link this PR to related issues or PRs.


zugao (commented):
DRY but this behemoth is unreviewable. Before deploying to prod please:
2) create one instance for each Exoscale DBaaS and ObjectBucket on your new version in lab cluster.
1) create one instance for each exocale DBaaS and object bucket on the current version. T

apis/exoscale/v1/iam_types.go
```diff
@@ -67,7 +68,7 @@ type IAMKeyObservation struct {
 	KeyID string `json:"keyID,omitempty"`
 
 	// RoleID is the observed unique ID as generated by exoscale.com.
-	RoleID string `json:"roleID,omitempty"`
+	RoleID exoscalesdk.UUID `json:"roleID,omitempty"`
```
zugao, line 71: This might not work with current versions, but hopefully I am wrong

wejdross: please @TheBigLee add tests for this case

operator/bucketcontroller/disconnect.go
```diff
@@ -0,0 +1,7 @@
+package bucketcontroller
+
+import "context"
+
+func (p *ProvisioningPipeline) Disconnect(ctx context.Context) error {
+	return nil
+}
```
wejdross, line 7: I assume we need that to satisfy an interface?

wejdross (commented):
just a few things from me, overall it looks good, just a reminder from my side:
1. Please test CRUD of all DB engines
2. Please test what happens to existing instances when You deploy new version

apis/exoscale/v1/iam_types.go
TheBigLee: I've added a test for this

TheBigLee:
>     1. create one instance for each Exoscale DBaaS and object bucket on the current version (master) then deploy your new version and see what happens. Try to update some of the objects.
> 
>     2. create one instance for each Exoscale DBaaS and ObjectBucket on your new version in lab cluster.
I've done both already on the Lab and both cases work as expected.

zugao (approved):
LGTM

operator/iamkeycontroller/observe.go
```diff
@@ -29,20 +25,20 @@ func (p *IAMKeyPipeline) Observe(ctx context.Context, mg resource.Managed) (mana
 	iamKey := fromManaged(mg)
 	// to manage state of new and old keys I need other variable, this is why this annotation is set
 	// otherwise observation fails for one of key types
-	if _, exists := iamKey.Annotations["newKeyType"]; exists {
```
Kidswiss, line 32: Won't this break legacy keys?

Kidswiss (approved):
It's way too much to review everything in detail.

There's also an update crossplane runtime, so please make sure to test everything with existing instances as well.

Also we might still have legacy iamkeys on some clusters, won't this change break them?

operator/iamkeycontroller/observe.go
zugao: I think Bigli removed them all, we only had legacy keys for ourselves, there were no customers using old keys.

TheBigLee: Yes it will, but I checked and there aren't any legacy keys in use for appcat

